PR #334, opened 2015-01-14
Disable toggle play key event option

henriklundgren:
I need a way to disable toggle event.


henriklundgren:
Sorry, found core.mediaControl.unbindKeyEvents();

PR #348, opened 2015-01-27
base: use default values for options

lewiscowper:
Hey there, I'm not sure how close I am, and I think there's more that I'm missing, but I thought I'd get your views on this fix for #346 while I'm early into ES6.

PR #350, opened 2015-01-28
base: use default values for options

### LINKED ISSUE #346: Use default values for arguments instead of || (or)
ES6 supports default values on methods so I think it's better to use it. Example (on [base_object.js](https://github.com/globocom/clappr/blob/master/src/base/base_object.js#L11-L14)):
```javascript
class BaseObject extends Events {
  constructor(options={}) {
    this.uniqueId = _.uniqueId('o')
    _.extend(this, _.pick(options, pluginOptions))
  }
}
```
instead of 
```javascript
class BaseObject extends Events {
  constructor(options) {
    this.uniqueId = _.uniqueId('o')
    options || (options = {})
    _.extend(this, _.pick(options, pluginOptions))
  }
}
```

lewiscowper:
I can't seem to re-open #348, but this PR should fix #346 as per comments there. :+1:

### DIFF
--- a/src/base/base_object.js
+++ b/src/base/base_object.js
@@ -9,9 +9,8 @@ var Events = require('events')
 var pluginOptions = ['container']
 
 class BaseObject extends Events {
-  constructor(options) {
+  constructor(options={}) {
     this.uniqueId = _.uniqueId('o')
-    options || (options = {})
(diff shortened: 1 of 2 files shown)


flavioribeiro:
thank you for this @lewiscowper, we hope to receive more contributions from you. :beers:

PR #361, opened 2015-02-15
Fixing gaTrackerName on GoogleAnalytics plugin

guilhermebruzzi:
When you used to pass a gaAccount parameter to player and didn't pass a gaTrackerName parameter, the default trackerName on GoogleAnalytics plugin was "undefined.".
I fixed that issue and added some tests to cover this GoogleAnalytics plugin's behaviour.
I also fixed the example on "README.md", because it was saying to pass a "gaTracker" parameter, but the correct parameter in the example would be "gaTrackerName".

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -15,6 +15,7 @@ Bernardo Camilo <bcvcamilo@gmail.com>
 Bruno Torres <me@brunotorr.es>
 Felipe Victor (Baiano) <felipevictor13@gmail.com>
 Flávio Ribeiro <email@flavioribeiro.com>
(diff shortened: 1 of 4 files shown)


flavioribeiro:
@guilhermebruzzi thank you for this, I would suggest you to avoid adding clappr.js and clappr.min.js on the pull request so I can merge it. Add yourself on CONTRIBUTORS also. Thank you again!

guilhermebruzzi:
@flavioribeiro Hi Flavio, i removed the commit with the files of dist and added another commit with my name on CONTRIBUTORS.

flavioribeiro:
thank you @guilhermebruzzi!

PR #365, opened 2015-02-23
Google Analytics: Fixing embed script sequence

celiofonseca:
Running embedScript just after setting analytics account information, fixing problem when sending events and 'window._gat' is already defined.


flavioribeiro:
thanks @celiofonseca

Issue #369, opened 2015-02-27
iOs full screen


irongomme:
I can't get full screen mode on iOs (7 and 8)

I tap the full screen button, but nothing happens ...

irongomme:
A new bug with full screen on iOs : 

If i have swtiched to fullscreen mode and then if i go back to original size, it doesn't remove the fullscreen css class from the player html tag. This is pretty logic, because iOs uses it's own interface on fullscreen mode, and buttons from this interface doesn't trigger clappr events.

In some case, this situation comes with troubles, because fullscreen class set à 100% height, so if the player is under a parent tag without fixed height, it is just not visible anymore.

I can fix it, and i will need this issue to be re-opened

PR #372, opened 2015-03-03
#369 add fullscreen support on iOs 8

irongomme:
I've got it working with this fix !

### DIFF
--- a/src/base/utils.js
+++ b/src/base/utils.js
@@ -53,13 +53,15 @@ var formatTime = function(time) {
 
 var Fullscreen = {
   isFullscreen: function() {
-    return document.webkitIsFullScreen || document.mozFullScreen || !!document.msFullscreenElement;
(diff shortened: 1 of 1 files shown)


src/base/utils.js
```diff
@@ -53,13 +53,15 @@ var formatTime = function(time) {
 
 var Fullscreen = {
   isFullscreen: function() {
-    return document.webkitIsFullScreen || document.mozFullScreen || !!document.msFullscreenElement;
+    return document.webkitFullscreenElement || document.webkitIsFullScreen || document.mozFullScreen || !!document.msFullscreenElement;
```
irongomme: I can split it like this : 

```
return (
      document.webkitFullscreenElement || 
      document.webkitIsFullScreen || 
      document.mozFullScreen || 
      !!document.msFullscreenElement
);
```

is this really usefull ?

flavioribeiro:
thanks @irongomme!

PR #374, opened 2015-03-03
Chromecast support

princ007:


thiagopnts:
I think this should be a external plugin. What do you guys think?
Looks like this increases ~7kb in the final bundle.

barbosa:
@princ007 This PR should not be integrated yet once the feature is still under development.

@thiagopnts I think there are pros and cons. Actually, this decision was not made yet. It was added here just for testing purpose. Perhaps, before all, we need to discuss what makes sense to be delivered as an external plugin or inside clappr's bundle. But yeah, let's talk about it in a near future :)

flavioribeiro:
Just my two cents. We're in a battle to decrease bundle size and chromecast support isn't a base feature like supporting mp4 or HLS playback. If we add chromecast as a core functionality, we'll need to add all the other OTT devices on it.

dayvson:
I understand the reason of why you guys want to test and experiment with Chromecast and decide to make it a core plugin, but I believe this decision should be re-evaluated. 

Mainly for two reasons, 
1 - To make this code work, it will require that a user have a Chromecast device, which is an enormous external dependency. I believe this reason should be strong enough to make the case of make Chromecast as an external plugin. 

2 - Also, there are no tests today about for this plugin; I believe Clappr core plugins and code base should have a high test coverage and be as lightweight as possible.

To put it into another perspective, this plugin as it is today does not add that much value for most of the users and it makes the codebase bigger. In a scenario where a user needs to Cast audio/videos in his application is perfectly reasonable to add an external plugin to enable the feature.

I can move it to an external plugin and send a pull request if you guys agree.

Issue #373, opened 2015-03-03
Centered play icon keeps its original size


irongomme:
The play icon on the center of the video is not visible because of its size of 4px

I've noticed it on : 
  - Chrome linux
  - Firefox linux
  - Safari iOs 7 iphone and iOs 8 ipad
  - Chrome Android 4.1 
  - Internet Android 4.1
  - IE 10 Surface

This is a css issue, i can take a look on it ... !

irongomme:
I've got it : 

**In facts, this happens when we use percents dimensions for the player height**

  - The poster plugin listen for resize and apply a font-size equal to the playerInfo height (height defined in the player instanciation)
  - When i set a 100% height, it set a 100% font-size and this is the problem !

I see in src/plugins/poster/poster.js (l.103) :
```
var height = PlayerInfo.currentSize ? PlayerInfo.currentSize.height : this.$el.height()
```

So, why can't we simply just put : 
```
var height = this.$el.height()
```

???

This works for me with this fix ...

thiagopnts:
Sorry for the delay, and thanks for investigating on this. By "player height 100%" you mean the `height` attribute on player initialization?

irongomme:
Yes, i need the player to be responsive

thiagopnts:
I see, this change was introduced by @towerz at https://github.com/clappr/clappr/commit/e364cd3db277122aefb53fe499dcc72cb65c61ef to keep the icon size proportional when changing player's size. Currently we don't support percent height/width, we need the `data-player` div to have a fixed size for reference to the inner elements. For adaptive size, you would need to use `.resize({width/height})`.

towerz:
@irongomme Please check whether the problem persists using the latest version (0.0.114). I believe this behavior has been fixed for percentage values.

irongomme:
@towerz 

:ok_hand: 

Great, it works as expected, and i don't have to maintain my own branch to have this behavior, thxs !

towerz:
@irongomme Anytime! I'm closing this issue then. If you find any other problem please reopen.

PR #376, opened 2015-03-04
#373 Fix play icon height with player height percents

irongomme:
I think $el.height() is enough and make percent height for the player able to display the big play button in a good height

### DIFF
--- a/src/plugins/poster/poster.js
+++ b/src/plugins/poster/poster.js
@@ -100,7 +100,7 @@ class PosterPlugin extends UIContainerPlugin {
 
   updateSize() {
     if (this.container.playback.name === 'html_img') return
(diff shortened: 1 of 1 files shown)


irongomme:
Need a fix for resize play button on player resize

thiagopnts:
Actually this is related to using percentage as height/width, which we don't support

irongomme:
@thiagopnts Do you plan to support it ?

thiagopnts:
@irongomme Not right now. The internal parts of the player currently depends on a fixed height/width from its "root" element. I would suggest you to manually set the player's height/width using the `.resize({width: width, height: height})` method from the player instance as you need.

Issue #375, opened 2015-03-04
Volume control from media_control not working on iOs


irongomme:
On iOs 8, increase, decrease or mute volume from the media_control has no effects, except for the ui rendering.

Volume can only be controlled by the native fullscreen video ... and when i search over internet : 

http://blog.millermedeiros.com/unsolved-html5-video-issues-on-ios/
//No way to control volume using JavaScript//

Thanks you Apple !

PR #377, opened 2015-03-04
#375 hide volume controls on ios devices

irongomme:
In facts, we cannot change how iOs works, so let's hide those controls on it !


flavioribeiro:
thanks @irongomme !

PR #422, opened 2015-04-20
#369 Fix coming back to original size from fullscreen mode on iOs

irongomme:


thiagopnts:
I tested on iOS and it works fine but looks like this broke the fullscreen in Safari(desktop), can you check it?

irongomme:
@thiagopnts you're right, i might be ok now !

thiagopnts:
@irongomme thanks!

PR #455, opened 2015-05-15
html5_video: dispatch buffering/bufferfull events when there's no enough buffer (fixes #26)

### LINKED ISSUE #26: Fix spinner events
Spinner is appearing only on the first rebuffer

flavioribeiro:
I've tested with chrome bandwidth throttling tool, it should be working fine.


towerz:
Thanks @flavioribeiro! :beers:

PR #465, opened 2015-05-22
Fixed extension regexes and added other hls mime type

### LINKED ISSUE #464: HLS on firefox
Hi,

I made a test page with a HLS VOD streaming from a Wowza Media Server :
http://91.121.17.70/clappr.html

The player is working great on Chrome / IE / Safari. But fail on firefox.

It seems that on Firefox only clappr load Player.swf instead of HLSPlayer.swf

tjenkinson:
Fixes #464


towerz:
Hi @tjenkinson, unfortunately this solution breaks using query strings in the player source. I ended using an implementation similar to what is already used in the HTML5 video playback.

Thanks for the help!

PR #489, opened 2015-06-12
Fixed watermark position for top-right top-left + small tweak

stanogurnik:
Fixing position for watermark when on position top-right , top-left + set max width on image inside wrapper + set relative watermark size based on player width,

Tested on all major browsers
Test logo collection: https://drive.google.com/file/d/0Bz9apPz75nEWbWZTZWsxY3U5ZkU/view?usp=sharing

### DIFF
--- a/src/plugins/watermark/public/watermark.css
+++ b/src/plugins/watermark/public/watermark.css
@@ -1,11 +1,12 @@
 [data-watermark] {
   position: absolute;
-  margin: 100px auto 0;
-  width: 70px;
+  width: 12%;
   text-align: center;
   z-index: 10;
-  margin-left: auto;
-  margin-right: auto;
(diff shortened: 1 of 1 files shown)


leandromoreira:
@stanogurnik thanks D:D !

It would be good if you'd rebase your code first. Also, I don't know if the original intent of this plugin was to have a fixed (70px) or a floating size (12%).

What do you think @flavioribeiro @towerz @bernardocamilo @thiagopnts @all ?

stanogurnik:
Maybe best solution is to set also min-width 70px (for too-small sized players) & max-width 200px (for fullscreen) ?

leandromoreira:
I agree with @stanogurnik can you rebase your code?

PR #490, opened 2015-06-15
Fixing Uint32Array is not woking on IE10 with canvas context

stanogurnik:
Fixing error page due to "Uint32Array was available to IE10, but it wasn't implemented in canvas"
http://stackoverflow.com/questions/26138234/array-buffer-ie10-typed-array-constructor-argument-is-invalid


towerz:
Thanks @stanogurnik!

PR #491, opened 2015-06-17
Fixed mixed content for ga.js

stanogurnik:
Fixing :  Mixed Content: The page at '{host}' was loaded over HTTPS, but requested an insecure script 'http://www.google-analytics.com/ga.js'. This request has been blocked; the content must be served over HTTPS.


leandromoreira:
Thanks a lot @stanogurnik  :) I would vote for us to use https given https://github.com/cdnjs/cdnjs/issues/4112 what do you think? do you see any problem by going https only @towerz @flavioribeiro @thiagopnts @bernardocamilo  ?

towerz:
@leandromoreira I think using the same protocol to load ga is better than using https only.

PR #505, opened 2015-07-26
utils: added days to formatTime function

derrod:
This is a change I made for my use case with a playlist longer than 24 hours (HLS based DVR).

At 27 hours the timeline will now display as `1:03:00:00` instead of `03:00:00`. This fixes the visual bug where the timeline starts at 0 after the first 24 hours.

Alternatively you could just display the full hours (i.e. `27:00:00`). However this seemed like a more elegant solution.


leandromoreira:
nice @derrod , can you try if you can squash the commits so it appears cleaner in git history: http://stackoverflow.com/questions/5667884/how-to-squash-commits-in-git-after-they-have-been-pushed

derrod:
God damn it. I have never done this in git before, I must have fucked something up, I'll close this pull-request and create a new one after I fix my fork.

leandromoreira:
:) @derrod tell me if you need any help

derrod:
@leandromoreira I created a new pull request #507.

I guess I should learn some git :P

PR #516, opened 2015-08-05
fix spinner_three_bounce plugin so that dots animate in sequence

tjenkinson:
Previously the first 2 dots animated at the same time, not sure if this was intentional?


towerz:
I think this was not intentional, but this way the animation seems better. Thanks @tjenkinson!

Issue #519, opened 2015-08-12
Player does not initialize as it fails to load a dependency that seems to have been taken off


lkkadiri:
GET http://cdn.clappr.io/latest/assets/HLSPlayer.swf?inline=1 request fails even with a specified version of clappr(0.1.7) is being used. Tried upgrading to '0.2.1' and the player does not initialize anymore. Is there any documentation on breaking changes? This affects all our environments. We are currently specifying clappr in our node application and requiring clappr.

konashark:
I entered a similar issue earlier today, but I found moving to 0.2.1 fixed the issue and worked without changes (my previous version was 0.1.1) and got rid of the dependency, so I closed the issue. Did you clear your browser cache after changing versions?

lkkadiri:
@konashark I have tried clearing my cache and it does not seem to work and i tried building the latest master locally to only find out the bin/hook.js is no more included in the project and the build seems to fail.

konashark:
Oh? I never build anything, I just copy the .js files out of 'dist' and drop them into my project. (BTW: I do not represent Clappr in any way, just trying to help since I had the same problem this morning and I was lucky that the update worked out of the box for us.) Good luck!

lkkadiri:
@konashark Thanks for the suggestions, still no luck for us here. I was trying to build as a last resort but i can only build the stable version which still uses that cdn dependency for the hls swf file.

thiagopnts:
@lkkadiri yes, we changed the way we build the player, I'll update update the README with the right instructions, now you should just `npm run build` or `npm run release`

lkkadiri:
Yeah we figured that out thanks @thiagopnts. Also opening a PR to address the playbackhls that still uses window.Clappr

towerz:
@lkkadiri clappr 0.1.x had a limitation that prevented us to lock the default base url for the assets to the current version. From 0.2.0 onward we'll not have this problem, unless the `baseUrl` parameter is explicitly set to point to `latest`.

Since we were in the process of changing the way the player is built, this ended breaking instances of 0.1.x without setting `baseUrl`. It's now fixed on 0.2.2, and using `latest` will not break loading assets anymore.

jfairley:
@towerz We're actually using CommonJS + Browserify to import clappr. For this reason we were surprised that a CDN is getting used at all (the swf is coming from CDN). Are there future plans to completely remove the depedency on the CDN when using CommonJS?

towerz:
[...]

PR #520, opened 2015-08-12
initialize missing global variable

jfairley:
referencing: https://github.com/clappr/clappr/issues/519

Now that the clappr player is no longer using `window` to store variables (https://github.com/clappr/clappr/blob/0.2.1/src/main.js#L21), `window.Clappr` is not getting initialized. It seems that the Flash HLS code was missed in this upgrade.

This may not be the most elegant solution, but the flash player still needs some sort of global hook for its callbacks. I will let y'all decide a better way to solve this, but here is one solution that is working for us locally.

### DIFF
--- a/src/playbacks/hls/hls.js
+++ b/src/playbacks/hls/hls.js
@@ -351,6 +351,9 @@ export default class HLS extends Playback {
   }
 
   createCallbacks() {
+    if (!window.Clappr) {
+      window.Clappr = {}
+    }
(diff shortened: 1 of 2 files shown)


jfairley:
btw, I'll start working on a unit test for this change....

lkkadiri:
:+1:

thiagopnts:
thanks @jfairley!

PR #525, opened 2015-08-18
Fix so that seek time doesn't flicker when hovering

tjenkinson:
Previously the `showTime` and `update` methods both queried `$el` for it's width, but if `$el` wasn't visible its width would be reported as 0 by the browser, resulting in the time then being shown in the wrong place. This change ensures that the element is always visible before the width queries are made, and prevents the time jumping left and right when hovering.

This also makes sure it doesn't wrap onto 2 lines when it reaches the far right of the player. Previously if it did wrap this would also cause the width to change and it would get a bit confused.

Also the previous code didn't, and this change currently doesn't listen to see if the `this.mediaControl.container.settings.seekEnabled` option changes, and then update appropriately. E.g. the user is hovering over the bar whilst `seekEnabled` is `true`, and then it flips to `false`. The time will still be visible until the cursor is moved. The reverse behaviour can happen when a HLS stream becomes longer than the minimum duration to DVR, causing DVR to become enabled.

Is there an event that is fired when settings change? If so I can have a look at adding this functionality in as well if you'd like?

### DIFF
--- a/src/components/seek_time/public/seek_time.scss
+++ b/src/components/seek_time/public/seek_time.scss
@@ -2,6 +2,7 @@
 
 .seek-time[data-seek-time] {
   position: absolute;
+  white-space: nowrap;
   width: auto;
   height: 20px;
   line-height: 20px;
--- a/src/components/seek_time/seek_time.js
+++ b/src/components/seek_time/seek_time.js
@@ -34,6 +34,12 @@ export default class SeekTime extends UIObject {
   }
 
   showTime(event) {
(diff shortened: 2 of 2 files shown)


towerz:
Thanks @tjenkinson! I ran into some of these issues, but didn't spend some time to investigate this problem and create an issue.

About the event triggered when settings are updated, there's the event Clappr.Events.CONTAINER_SETTINGSUPDATE you could listen to.

PR #528, opened 2015-08-21
SeekTime Now Handles When "seekEnabled" setting changes

tjenkinson:
It doesn't yet but hopefully will by the end of this pull request!

This is what I have so far, it's basically the same as before just refactored slightly.

I have some questions:
1. How do I get access to the container to listen for the settings changed event? Does it need to be a UICorePlugin to get access to this?
2. Where in the code does the duration value get added and updated? It feels like this should be something that is controlled here as well?
 
image

Thanks


leandromoreira:
@tjenkinson thanks so much, you always help a lot. :smile:

Some notes though: 

*  instead of doing `var self = this` use arrow function:
```javascript
    this.listenTo(this.mediaControl, Events.MEDIACONTROL_MOUSEMOVE_SEEKBAR, (event)  =>  {
      this.hoveringOverSeekBar = true
      this.calculateHoverPosition(event)
      this.update()
    })
```
*  avoid `;`


> "1. How do I get access to the container to listen for the settings changed event? Does it need to be a UICorePlugin to get access to this?"

It doesn't need to be a `UICorePlugin` and I think you can achieve this by listening for settings update.
```javascript
this.listenTo(this.mediaControl.container, Events.CONTAINER_SETTINGSUPDATE, () => {})
```

> "2. Where in the code does the duration value get added and updated? It feels like this should be something that is controlled here as well?"

You usually can listen to container time update or progress.
```javascript
Events.CONTAINER_TIMEUPDATE
Events.CONTAINER_PROGRESS
```

Both cases you can see examples through the code and I feel that we also need to add more test cases for code we add.

leandromoreira:
@tjenkinson I beg your pardon I didn't answer you. (I read fast and :hankey:)

> 2. Where in the code does the duration value get added and updated? It feels like this should be something that is controlled here as well? 

It depends on the playback: 

* For html5_video it listens to [durationchange](https://github.com/clappr/clappr/blob/master/src/playbacks/html5_video/html5_video.js#L37) event from video tag and then it know that it can query `$videotag.duration`.

* For flashls it does listen to [some events on flash](https://github.com/clappr/clappr/blob/master/src/playbacks/hls/flashls_events.js#L19) and then [it updates the duration](https://github.com/clappr/clappr/blob/master/src/playbacks/hls/hls.js#L359) (while you're playing a vod you have a fixed *duration* but with live different.

leandromoreira:
@tjenkinson sorry once more :(

Now I could understand and I agree with you but maybe there is a reason for this component, what you think @towerz @thiagopnts @flavioribeiro

towerz:
@leandromoreira @tjenkinson Actually, the DvrControls plugin was added later on the player development and we didn't revisit these plugins after that.

The reason behind DvrControls handling the duration value is that SeekTime is not actually a plugin, it's a part of MediaControl that we separated from the main file to keep it better organized. We wanted to make available for plugins to alter the appearance/structure of the media controls, and this manipulation is one of the examples we wanted to show. 

In fact, we didn't want to make MediaControl know what DVR is. The plugin for handling DVR should be the only one to know what dvr is and change the controls accordingly.

@tjenkinson's comment pointed out a failure in that. We should either remove DVR state handling (CONTAINER_PLAYBACKDVRSTATECHANGED event) from MediaControl or we follow his suggestion to handle duration directly in SeekTime. I'm opposed to the latter because we would need to add VoD/live/DVR checks to SeekTime, instead of giving it a single responsibility of displaying the current time of the hovering position on the seek bar.

towerz:
@tjenkinson due to a limitation in our build process, SeekTime could not be included in the same folder as MediaControl. There were conflicts when loading html/css templates. Now that we use webpack for managing these assets, we don't have this limitation anymore.

About the events, you are right. Both events cause MediaControl to re-render, so there's no need to listen for a settings change event.

leandromoreira:
;) no problem @tjenkinson you always help a lot :smile: 

>In MediaControl there is a destroy() method. Where is this called from? I'm a bit confused as to where this fits in with the system. Also shouldn't the stuff it does also happen if the user was to call remove()?

I think the `Core` is responsible for the life cycle of `MediaControl` (alongside with `Container` and all the plugins which also should have `destroy` method), therefore when you `destroy` the core it'll destroy also all the plugins and the media control.

You can see that [here](https://github.com/clappr/clappr/blob/master/src/components/core/core.js#L149).

```javascript
  destroy() {
    this.disableResizeObserver()
    this.containers.forEach((container) => container.destroy())
    this.plugins.forEach((plugin) => plugin.destroy())
    this.$el.remove()
    this.mediaControl.destroy()
    $(document).unbind('fullscreenchange')
    $(document).unbind('MSFullscreenChange')
    $(document).unbind('mozfullscreenchange')
}
```

and you will destroy the `Core` when you [destroy your player](http://clappr.github.io/classes/Player.html#method_destroy).

> Also shouldn't the stuff it does also happen if the user was to call remove()?

I think the API for plugins is the `destroy` method, although I see that [`UIObject` employs the `remove` API instead](http://clappr.github.io/classes/UIObject.html#method_remove)... anyway I think @towerz @thiagopnts @flavioribeiro can help us out here, in these two questions.

* In MediaControl there is a destroy() method. Where is this called from? I'm a bit confused as to where this fits in with the system. Also shouldn't the stuff it does also happen if the user was to call remove()?
* Also shouldn't the stuff it does also happen if the user was to call remove()?

leandromoreira:
@tjenkinson yeah, you're right but since I'm kinda new I want the others to check if we are correct on this too, then we can fix this possible **bug**.

While the super method `remove` does that (I mean it calls `stopListening` and `undelegatingEvents`, which stops listening all the `events`) that's why I suppose we might need to call within `mediaControl#destroy`

```javascript
  remove() {
    this.$el.remove()
    this.stopListening()
    this.undelegateEvents()
    return this
  }
```

towerz:
@leandromoreira @tjenkinson The `destroy()` api was created to remove event handlers, release resources and remove the element from its parent. 

This was made before creating the `remove()` method for the `UIObject`, which should replace `destroy` as the method responsible for freeing resources and event handlers from the object, thus removing the responsibility from the plugin creators to take care of this. 

Apparently, the plugin base classes have been changed to call `remove` from their `destroy` methods, but some `UIObject`s have not been changed to do it, so it's actually a bug.

About the other two questions:
- The responsible for the lifecycle of `MediaControl` is indeed the `Core` object, @leandromoreira is correct in that.
- The user should not call `remove` at all, since it's not exposed in the `Player` API.

To remove doubts about this, I think we should remove the `destroy` methods, using only `remove` instead.

leandromoreira:
I'm cool with that, once again thanks so much @tjenkinson :clap: :clap: :clap:

towerz:
Thanks again, @tjenkinson.

PR #536, opened 2015-08-28
three spinner bounce fix to make sure only one show timeout created

tjenkinson:
Previously if `onBuffering` was called twice (which was happening for me sometimes when playing a vod hls stream), 2 timeouts would be created, and then if `onStop()` was called within the timeout interval, only the first of the timeouts would get cleared. This would then cause the loading indicator to reappear.

This should fix this but I haven't built and tested it yet as having some issues npm installing due to windows path length issues.


towerz:
Merged, thanks @tjenkinson!

Issue #460, opened 2015-05-19
Show a "Please install flash" message when flash required and not installed


tjenkinson:
At the moment in a fresh install of firefox without flash I get this:

![image](https://cloud.githubusercontent.com/assets/3259993/7701288/a13d94b2-fe1d-11e4-8bd6-dba91e834dd4.png)

After installing flash it works fine. Would be nice if the user was informed that flash was the issue instead of being told to switch browser. When I have time I can take a look at this if you want?

towerz:
@tjenkinson Sure, you could take a look at this. I suggest creating a plugin for this, since the No-Op playback is only displayed when no other option is found, and not every browser needs flash for the playbacks.

This "error" screen is not entirely wrong, you just need a plugin to check whether installing flash would allow the player to find a playback for the video and inform the user to install it.

towerz:
Hi @tjenkinson, any progress with this?

tjenkinson:
No sorry I'm pretty busy with other stuff at the moment. I'll comment here if I get time.

tjenkinson:
So I'm thinking a method could be added to Playbacks 'flashRequired()' which would return a boolean.

For everything where this may return true there should also be a 'canPlayWithFlash()' method which is the same as 'canPlay' but assumes flash is present.

Then there could be another playback called 'no_flash', which would check the other playbacks to see if they 'canPlayWithFlash()' and if it gets a true report true from its 'canPlay()'.

Sound ok?

tjenkinson:
Also chrome is now defaulting to some form of 'click to play' for plugins so this might become more important. Think it might automatically allow plugins from the same domain but haven't looked into it in depth yet.

towerz:
@tjenkinson I thinks there's only one problem with your idea. It creates a playback dependency on other playbacks.

The 'no_flash' playback checking other playbacks breaks the idea of having atomic playbacks. Playbacks shouldn't be aware of the others, unless it's a composition or inheritance relationship.

I think you could improve your idea using a core plugin instead of a 'no_flash' playback. The core plugin could listen to a `CONTAINER_READY` event, and check whether the playback is a 'no_op' playback. If it's a 'no_op', it could check whether the Playbacks respond true to `flashRequired` and `canPlayWithFlash` the source. This way you don't break the architecture, as well as avoid creating a meta-playback to check whether the others are flash-dependent.

What do you think?

tjenkinson:
[...]

PR #559, opened 2015-09-23
Added a "NoFlash" Playback

LGprodigy:
This will detect if the browser requires flash to playback the required
media when there is no HTML5/JS support. If flash isn't installed on the
browser, then a message pertaining to adobe flash will be displayed
instead of "Your browser does not support the playback of this video.
Try to use a different browser."!

### DIFF
--- a/src/components/loader/loader.js
+++ b/src/components/loader/loader.js
@@ -12,6 +12,7 @@ import FlashVideoPlayback from 'playbacks/flash'
 import HTML5AudioPlayback from 'playbacks/html5_audio'
 import HLSVideoPlayback from 'playbacks/hls'
 import HTMLImgPlayback from 'playbacks/html_img'
+import NoFlash from 'playbacks/no_flash'
 import NoOp from 'playbacks/no_op'
 
 /* Container Plugins */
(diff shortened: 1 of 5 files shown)


LGprodigy:
This will display when flash is missing:
https://gyazo.com/567c809e47c1d6bfb147fa6a872e0aa6

towerz:
Hi @LGprodigy, thanks for your PR, but there's some problems with your implementation that @tjenkinson and I have discussed previously on #460.

Knowing whether Flash is required for playback is indeed one of the responsibilities of a Flash-based playback, but not deciding whether or not to tell the user to install Flash. The sole responsibility of a playback is determine if the source received is supported, and accept or reject playing the video accoridngly. This is done by the `Loader` for all playback objects through the `canPlay` method.

The only exception to this is the `no_op` playback, which is a marker playback to indicate that the source is not supported in this browser. That's why there's a generic message informing the user to try on another browser.

The right way to implement a check for Flash is checking for all Flash based playbacks if they would be able to play the video if Flash was installed. Since this require knowledge of all playback plugins registered with the `Loader`, a `core` plugin would be a better option.

This plugin could be either a `UICorePlugin`, if it needs to display a custom message screen, or a `CorePlugin` if the only need for customization is the message shown in the `no_op` screen. In the latter case, the `no_op` plugin would need the addition of a setter to customize the message displayed.

Issue #538, opened 2015-08-31
player.getDuration() always returns Infinity on Safari native html5 player


dviramontes:
Hi 
we are trying to implement a set of custom player controls using Clappr's API 
and we've noticed that while streaming a video in Safari(v8.0.8), the html5 player
always returns: 

**Infinity** for `player.getDuration();`
and
**1** for `player.getCurrentTime();`

making it impossible to calculate a seek value.
(even if we pause the video, `player.core.getPlaybackType()`
always returns **live**).

We also noticed that using Clappr's own controls we get the same
behavior. Wondering if you all had come across this limitation before
and/or possible a way around it ?

Thanks, 
cc: @jfairley @lkkadiri

leandromoreira:
Hi @dviramontes 

I'm using a **Safari 8.0.8** and then I tried a `mp4` file at http://cdn.clappr.io and it worked:

> p.core.getPlaybackType()
< "vod"
> p.core.getCurrentPlayback().getPlaybackType()
< "vod"
> p.core.getCurrentPlayback().getCurrentTime()
< 4.237111111111111
> p.core.getCurrentPlayback().getDuration()
< 36.44

So I changed the `source` to `"http://www.streambox.fr/playlists/test_001/stream.m3u8"` and it still works:

> p.core.getCurrentPlayback().getPlaybackType()
< "vod"
> p.core.getCurrentPlayback().getCurrentTime()
< 3.2469333333333332
> p.core.getCurrentPlayback().getDuration()
< 510
> p.core.getCurrentPlayback().getCurrentTime()
< 7.611636596
> p.core.getCurrentPlayback().getCurrentTime()
< 9.134704286

I also tested with `player.getCurrentTime()` and I still couldn't get `Infinity`, although I used the instance `p` you should have the same behavior while using the right instance of the Clappr.

can you provide more detail so we can reproduce it ?

dviramontes:
I tried using the above mentioned functions
`p.core.getCurrentPlayback()` but it seems that 
`getCurrentPlayback()` is not longer defined (Using clappr 0.2.8).

but clappr does have something called `getCurrentContainer()`
so i tried that and I got: 
![screen shot 2015-09-01 at 11 23 12 am](https://cloud.githubusercontent.com/assets/144226/9611848/416f4bf0-509f-11e5-9440-2d4da1dafe17.png)

by the way,  we are using a `.m3u8` file as well and in our case
while are playing a playlist 
[...]

PR #561, opened 2015-09-23
HLS support on JS (closes #555)

### LINKED ISSUE #555: Change HLS Playback plugin to play in pure javascript
Today [hlsjs](https://github.com/dailymotion/hls.js) became open source. Since most browsers are preventing the use of flash and we already have a [initiative](http://github.com/flavioribeiro/clappr-hls-js-plugin) to use hlsjs on clappr, it would be great to go in the direction to remove flash from the built-in plugin.

flavioribeiro:
- [x] Support live videos
- [x] Support on demand videos
- [x] Support DVR


thiagopnts:
@houndci shut up

leandromoreira:
@thiagopnts we should fire him :+1: I'm in favor or maybe we instruct him to ignore `Line is too long.`

dviramontes:
*Apologies from jumping the gun here...*

In testing this PR (functional testing in latest stable Chrome and Safari), 
I saw that the playback does in fact default to a native `<video/>` player which is great! 

But I also saw that:
`p.core.mediaControl.container.getCurrentTime() ` returns **1** while in **live** mode.

I'm wondering if the **flashls.js** implementation used here also suffers from the same defects that were discussed in #538 

Maybe more of a problem with the spec?

flavioribeiro:
@dviramontes `hls.js` changes the TS video packets to be playable on a buffer by the `video` tag using MSE. This way, some of the method calls to the tag directly didn't reflect the current state of the playback. Actually, that's why this PR was not merged yet, I need to find the best way to implement the DVR behavior.

jfairley:
@flavioribeiro, I'm curious. How's this going? We're anxiously waiting. :smiley:

flavioribeiro:
hey @jfairley sorry, just moved to NY and had no time to finish this. I will try to finish it by this week.

src/playbacks/flashls/flashls.js
```diff
@@ -0,0 +1,482 @@
...
+    super.render()
+    this.createCallbacks()
+    return this
+  }
+}
+
+FlasHLS.canPlay = function(resource, mimeType) {
+  var resourceParts = resource.split('?')[0].match(/.*\.(.*)$/) || []
+  return Browser.hasFlash &&
+        ((resourceParts.length > 1 && resourceParts[1] == "m3u8") ||
```
tjenkinson: I guess hound might actually have a point here :P

PR #578, opened 2015-10-09
fix up this.settings attribution to avoid passing by reference

dayvson:
Also removing unnecessary semi-colon.


flavioribeiro:
:metal:

Issue #558, opened 2015-09-22
Opera browser does not support the playback


LGprodigy:
Hello,

My organization runs a very large platform worldwide. I have come across your video player. I want to implement it in our next release. After extensive testing the last few days, I keep coming across this issue in Opera browser https://gyazo.com/9fee3b3f2d2149ae21f6a4af91b3525d It is happening with any HLS stream I have tried. I thought maybe the builds we have been creating is causing the issues... but I have tried it with vanilla clappr as well. Is this a common problem? Am I possibly missing something? Open for recommendations and so forth. Thank you for your time!

tjenkinson:
Is the flash plugin installed and enabled? It is currently required for HLS in browsers that don't support HLS natively.

This should be [changing in the future though](https://github.com/clappr/clappr/issues/555) :)

LGprodigy:
I literally just answered my own question when you answered it. Thank you. Maybe a feature enhancement to think about... when using a flash wrapper alert the user the same way it displayed the error mentioned above. Something along the lines "Your browser requires flash player to play this media!". Just tossing that idea out there. I really am glad to hear hlsjs is open. Maybe scratch this idea! :D 

Thank you very much for your input!

tjenkinson:
No problem. I had the same idea a while ago and there is an issue here with more information https://github.com/clappr/clappr/issues/460.

I just haven't had time yet to implement it.

LGprodigy:
Well, if you guys are open to pull requests. Our team has no problem contributing!

tjenkinson:
I'm sure @towerz would be happy if you wanted to take a look at that issue. :)

LGprodigy:
We have created and tested an instance of missing flash when required. It works, we will be creating a pull request shortly for it.

LGprodigy:
#559
https://gyazo.com/567c809e47c1d6bfb147fa6a872e0aa6

flavioribeiro:
hey @LGprodigy, I made some tests with hlsjs running on opera and it seems to be working fine. I hope the guys will look at the #583 and merge it asap so this issue will be automatically fixed. I'll let it open to remember you to try before the merge and close or comment if you still find that it is an issue.

flavioribeiro:
I've tested clappr from cdn.clappr.io with some hls sources that I have and it's working fine here on my mac. I'm gonna close this since I don't think it's an issue anymore but feel free to reopen if you think that makes sense.

PR #583, opened 2015-10-13
HLS support on JS (closes #555) 

### LINKED ISSUE #555: Change HLS Playback plugin to play in pure javascript

flavioribeiro:
refs #561, hound messed up the PR with tons of alerts


flavioribeiro:
can anyone review this PR before we merge it? I made some smoke tests using [this](https://gist.github.com/flavioribeiro/49162ff16f916d0185a0) html and it seems to be working fine.

jfairley:
@flavioribeiro, unless someone else gives the thumbs up sooner, @dviramontes or I will pull it down and give it a test drive tomorrow at the office.

flavioribeiro:
great @jfairley feel free to do it!

jfairley:
Here's the config I'm running, modified from the default test page. You can't see our `.m3u8`, but there's nothing special about it.
```javascript
  source: 'http://10.0.1.69/media/hlsLiveTest01/stream/hls.m3u8',
  plugins: { playback: [Clappr.HLS] },
  baseUrl: '/latest',
  mute: true,
  height: 360,
  width: 640
});
```

## Chrome (47.0.2526.8 dev)
Everything looks great in Chrome, so it's definitely the test bed.

In live mode, the play head location is a little funky, but knowing how HLS works, I completely understand why it looks they way it does. I'm happy to see that we're getting accurate values for duration and currentTime. :+1: We both need to do some tweaking in our seek bars to make things look prettier.

## Opera (32.0)
I see the link to #558 about this not working in Opera. For me in Opera, the player looks every bit as good as it does in Chrome.

I did see one glitch in frame stepping, but it seems like the buffer wasn't loaded. When I quickly played/paused, I could step.

## Firefox (42.0b5)
It's funky in Firefox, but I would expect so, since FF doesn't support HLS. I was surprised that it plays and the audio is fine. It's only the video frame that FF fails to render correctly. NBD. Use Flash in FF.

About that... I tried declaring both playbacks, but it still tried to use HLS. I suppose we'll need to determine the browser at runtime and apply the correct plugins as necessary. Is that what you would expect?
```javascript
// does not work :(
plugins: { playback: [ Clappr.HLS, Clappr.FlasHLS ] }

// what I'll do instead
let playback = isFirefox ? Clappr.FlasHLS : Clappr.HLS
plugins: { playback: [ playback ] }
```

## Safari (9.0)
In Safari, when I declare the HLS plugin, my CPU pegs at 100%, and I get the spinning beach ball (OSX). Without the plugin (defaulting to native), everything runs fine. Of course with no plugin, the live mode currentTime / duration are still 1 / ∞ (or vice versa, I don't remember).

In fact, something nice that I found is that when no plugin is applied, Clappr correctly decides to use native in Safari, while it correctly decides to use HLS when in Chrome and Opera. :clap: Unfortunately, FF also tries to use HLS, too, so I'll need to keep my switch to force FF to use FlasHLS.

flavioribeiro:
wow @jfairley thanks for all the tests! I guess that the way to go is avoid hlsjs in ffox and safari for now.

dviramontes:
Nice testing indeed! :+1:

leandromoreira:
thanks @flavioribeiro :clap:  it's a great milestone for Clappr

towerz:
Thanks @flavioribeiro, excelent work!!! :beers:

jfairley:
woo hoo :beers:

mangui:
:beers: !!!

dviramontes:
:metal:

leandromoreira:
* Firefox 42 will enable MSE by default
* There is no whitelist anymore ; we won't be adding new sites to the whitelist

http://www.ghacks.net/2015/08/06/mozilla-to-enable-media-source-extensions-mse-for-all-websites-in-firefox-42/

https://bugzilla.mozilla.org/show_bug.cgi?id=778617

PR #590, opened 2015-10-17
hls playback fix so loads correct source

tjenkinson:
Previously was looking directly at 'source' in the options object not the resolved one.


leandromoreira:
Thanks @tjenkinson

PR #592, opened 2015-10-17
Fix so that use of hlsjs doesn't make clappr incompatible with requirejs and commonjs

tjenkinson:
This makes sure that hls.js doesn't register itself with requirejs and commonjs. Without this the built clappr file will first define itself as an anonymous module, then later on where hlsjs is inserted hlsjs would also try to define itself resulting in this error: http://requirejs.org/docs/errors.html#mismatch

The imports loader plugin will inject a new line to the start of hlsjs which overrides `define` to `false` therefore removing the issue. 

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,6 +37,7 @@
     "file-loader": "^0.8.4",
     "gulp": "^3.8.1",
     "html-loader": "^0.3.0",
+    "imports-loader": "^0.6.4",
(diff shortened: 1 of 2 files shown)


flavioribeiro:
@tjenkinson this pr still makes sense? if not, can you close it?

Issue #593, opened 2015-10-17
Option to disable right click on video tag


tjenkinson:
Would be nice if there was a config option which would disable the context menu on the video tag, when a video tag is used.

tjenkinson:
If you think this should be an option I've made a quick PR :)

PR #597, opened 2015-10-18
Added 'disableVideoTagContextMenu' option

tjenkinson:
Closes clappr/clappr#593


leandromoreira:
thanks again @tjenkinson can you also write the doc about this? `disableVideoTagContextMenu`

* one in here https://github.com/clappr/clappr/blob/master/src/components/player.js#L84
* and another at README like this one https://github.com/clappr/clappr/blob/master/README.md#loop (even though we might hold this PR for release date)

PR #602, opened 2015-10-21
Add tests for loader object and check if external plugins were placed on right array

### LINKED ISSUE #584: A check should be performed to make sure plugin is of correct type when loaded
E.g currently if you accidentally put a UiCore plugin as a 'container' plugin in the 'plugins' object in the 'options' object there are no errors.

Would be nice if there was an exception thrown.

flavioribeiro:
this pull request should close #584


flavioribeiro:
since the plugins are extending the base ones, they will automatically have the `type` attr, don't need to rebuilt them

flavioribeiro:
btw, if u guys think it makes sense I'll use the function and enhance the error message

flavioribeiro:
and fix travis :hankey:

PR #603, opened 2015-10-21
player: avoid duplicated code

leandromoreira:
Since we have `getPlugin(name)` at [Loader](https://github.com/clappr/clappr/blob/master/src/components/loader/loader.js#L75) and [Player](https://github.com/clappr/clappr/blob/master/src/components/player.js#L294) I think it's best to reuse this code.

### DIFF
--- a/src/components/player.js
+++ b/src/components/player.js
@@ -292,10 +292,7 @@ export default class Player extends BaseObject {
    * ```
    */
   getPlugin(name) {
-    var plugins = this.core.plugins.concat(this.core.mediaControl.container.plugins);
-    return find(plugins, function(plugin) {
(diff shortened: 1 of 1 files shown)


flavioribeiro:
it seems there's more `getPlugin()` functions on [container](https://github.com/clappr/clappr/blob/master/src/components/container/container.js#L293-L295) and [core](https://github.com/clappr/clappr/blob/master/src/components/core/core.js#L138-L140) not to mention `hasPlugin()`, `addPlugin()` functions with same code.. should we improve that also?

leandromoreira:
my bad @flavioribeiro both, container and player handles only container **instances** while the Loader return the classes.

PR #605, opened 2015-10-21
Use $.extend instead of lodash.assign (closes #579)

### LINKED ISSUE #579: Use assign() or $.extend() to create objects with new references
I saw some places in the code where we use `$.extend()` and `assign()` for the same purpose. It would be better if we use only one.

dayvson:


flavioribeiro:
@dayvson thanks! :beers:

PR #621, opened 2015-11-01
Fallback drag cursor now included as asset and volume drag now works same way as other media control drags

### LINKED ISSUE #613: Remove volume bar drag cursor external dependency
Just spotted that the hand cursor which is used when changing the volume is picked up from google. https://github.com/clappr/clappr/blob/77e016b3f416c06857db6b8789c7efd1eb85340c/src/components/media_control/media_control.js#L166

Any reason why this couldn't become an asset just like the fonts and flash players? The URL looks like it might be one that google could easily change.

tjenkinson:
Closes #613


flavioribeiro:
For some reason the media control couldn't find the cursor on travis build. Can you take a look on it before we merge this pr?

flavioribeiro:
here's the steps that travis do https://github.com/clappr/clappr/blob/master/.travis.yml

flavioribeiro:
thanks man

PR #623, opened 2015-11-02
Standardilizes *_PROGRESS event parameter

leandromoreira:
*This one is a little bit trick*, I was looking at *_PROGRESS events and noticed that we could provide a standard parameter to the progress event. 

Basically I think of `{start: start, current: current, total: total}` as these names pass the feeling of progress but maybe these are not the best names!?

## Just to put in context

It was:

```javascript
//container
this.trigger(Events.CONTAINER_PROGRESS, startPosition, endPosition, duration, this.name);
//flash
this.trigger(Events.PLAYBACK_PROGRESS, 0, this.el.getBytesLoaded(), this.el.getBytesTotal(), this.name)
/...
```

To become:
```javascript
//container
this.trigger(Events.CONTAINER_PROGRESS, progress)
//flash
this.trigger(Events.PLAYBACK_PROGRESS, {start: 0, current: this.el.getBytesLoaded(), total: this.el.getBytesTotal()})
//...
```

## Tests

I ran it on **Chrome (MacOS)** with two VOD's : **mp4** and **hls**. 

## Questions

I also ged rid off `this.name` parameter, can't we just add it within trigger itself?
Do we need `start` (`starPosition`) for this event?

### DIFF
--- a/src/base/events.js
+++ b/src/base/events.js
@@ -255,6 +255,19 @@ Events.PLAYER_ERROR = 'error'
 Events.PLAYER_TIMEUPDATE = 'timeupdate'
 
 // Playback Events
+/**
+ * Fired when the browser loads more of the media file at playback level
+ *
+ * @event PLAYBACK_PROGRESS
+ * @param {Object} progress Data
+ * progress object
+ * @param {Number} [progress.start]
+ * start progress
+ * @param {Number} [progress.current]
+ * current progress
+ * @param {Number} [progress.total]
+ * total progress
+ */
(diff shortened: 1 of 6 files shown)


leandromoreira:
Also, what's the difference between *_PROGRESS and *_TIMEUPDATE and main usage?

flavioribeiro:
Progress events accounts for media loading/buffering. Time update is dispatched when currentTime is updated, so it's related to the playback position

leandromoreira:
I'll see this after.

PR #624, opened 2015-11-02
Introduces high definition update event parameter

leandromoreira:
## Questions

I saw that [`seek`](https://github.com/clappr/clappr/blob/master/src/playbacks/flashls/flashls.js#L417) at flashls is triggering **high definition update** event and I removed it. Is there any reason for this?

## Tests
I couldn't test this, in a sense of show/hide('ing) HD icon. Although I played mp4 and hls with no problems.

### DIFF
--- a/.hound.yml
+++ b/.hound.yml
@@ -1,3 +1,0 @@
-java_script:
-  enabled: true
-  config_file: .jshintrc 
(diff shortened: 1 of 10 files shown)


leandromoreira:
I'm gonna rework this one (rebase, git fetch, git robot, git anything)

PR #625, opened 2015-11-02
Changes bitrate event parameter

leandromoreira:
Very inspired by [Shaka](https://shaka-player-demo.appspot.com/docs/shaka.media.IStream.html#event:AdaptationEvent) I propose that we follow this  API for bitrate event parameter:

```javascript
{bitrate,bandwidth,width,height,level}
```

## Questions

I'm not 100% sure about `level`, do we need that? it looks pretty coupled to HLS and maybe not useful outside this context.


flavioribeiro:
lgtm

PR #630, opened 2015-11-02
Improve HTML5Audio class to extend HTML5Video and remove unnecessary methods.

dayvson:


src/playbacks/html5_audio/html5_audio.js
```diff
@@ -5,158 +5,32 @@
 import Playback from 'base/playback'
```
tjenkinson: Think this can be removed now?

src/playbacks/html5_audio/html5_audio.js
```diff
@@ -5,158 +5,32 @@
...
   }
-
+  
   getPlaybackType() {
-    return [0, undefined, Infinity].indexOf(this.el.duration) >= 0 ? 'live' : 'aod'
+    return 'aod'
   }
 
   stalled() {
-    if (this.getPlaybackType() === 'vod' && this.el.readyState < this.el.HAVE_FUTURE_DATA) {
```
tjenkinson: Not sure why `this.getPlaybackType() === 'vod'` used to be there? Wouldn't it always have returned 'live' or 'aod'?? (line 64)

dayvson: I think it was a wrong copy and past from HTML5Video file. As far as I know the live support for audio on HTML5 will always play the audio using the tag video when the stream is an M3U8 with audio only. I did some tests and it is working fine.

flavioribeiro:
thanks @dayvson!

PR #643, opened 2015-11-07
no_op: show error message on browser's current language fixes #363

### LINKED ISSUE #363: NoOp Playback: Show error message on browser's current language

dayvson:
Languages available so far English, Spanish, Portuguese

### DIFF
--- a/src/playbacks/no_op/no_op.js
+++ b/src/playbacks/no_op/no_op.js
@@ -14,13 +14,24 @@ export default class NoOp extends Playback {
     return {'data-no-op': ''}
   }
 
+  getNoOpMessage(){
+    var messages = {}
+    messages['en'] = "Your browser does not support the playback of this video. Try to use a different browser."
+    messages['es'] = "Su navegador no soporta la reproducción de un video. Trate de usar un navegador diferente."
+    messages['pt'] = "Seu navegador não supporta a reprodução deste video. Tente usar um outro navegador."
+    messages['en-US'] = messages['en']
+    messages['es-419'] = messages['es']
+    messages['pt-BR'] = messages['pt']
+    return messages[window.navigator.language]
+  }
+
   constructor(options) {
     super(options);
   }
 
   render() {
     var style = Styler.getStyleFor(noOpStyle);
-    this.$el.html(this.template())
+    this.$el.html(this.template({message:this.getNoOpMessage()}))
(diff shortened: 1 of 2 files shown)


src/playbacks/no_op/no_op.js
```diff
@@ -14,13 +14,24 @@ export default class NoOp extends Playback {
...
 
+  getNoOpMessage(){
+    var messages = {}
+    messages['en'] = "Your browser does not support the playback of this video. Try to use a different browser."
+    messages['es'] = "Su navegador no soporta la reproducción de un video. Trate de usar un navegador diferente."
+    messages['pt'] = "Seu navegador não supporta a reprodução deste video. Tente usar um outro navegador."
+    messages['en-US'] = messages['en']
+    messages['es-419'] = messages['es']
+    messages['pt-BR'] = messages['pt']
+    return messages[window.navigator.language]
```
leandromoreira: Thanks @dayvson  :+1: , the only thing I would change would be a more safe about `window.navigator.language` since it seems that it's [not 100% supported by all browser](https://developer.mozilla.org/en-US/docs/Web/API/NavigatorLanguage/language)

We could start to [make use of polyfill!](http://gu.illau.me/posts/follow-up-on-navigator-languages/#polyfill) but instead of spreading this code we could start an object that is called at Core.

```javascript
class ClapprPolyfill {
   installAll() {
       this._installNavigator()
   }
   
  _installNavigator() {
    if (!window.navigator) { window.navigator = {} }
    if (!window.navigator.language) { window.navigator.language = 'en-US' }
   }
}
```

What do you think @flavioribeiro @towerz @thiagopnts @bernardocamilo @tjenkinson ?

dayvson: nice catch @leandromoreira I created the installNavigator function inside the NoOP I think this should be fine.

leandromoreira:
:beers:

towerz:
I'm inclined to agree with @tjenkinson. I don't think setting globals is a good way to go.

Perhaps we could use something like what we have in the utils package, where we have helper functions that use the browser built-in globals or fall back gracefully to a polyfill-like implementation.

dayvson:
that makes sense, I will send another pull request fixing it.

dayvson:
https://github.com/clappr/clappr/pull/647

PR #678, opened 2015-11-16
Fixes in HLS playback so that playable content start point is now correct

### LINKED ISSUE #668: artificial time drift in live mode for native HLS (0.2.18)
This is a weird one. I've tracked down the source, but I'm not sure of the correct way to fix it.

My scenario is that I have a live HLS feed. I seek back so that I'm in DVR mode. Every 10 seconds, the current time reported from `PLAYER_TIMEUPDATE` jumps 10 seconds. Watching the video proves that the video is not actually jumping 10 seconds forward, but the current time value artificially gets 10 seconds larger at regular intervals.

After some investigations, I see that this is definitely tied to when new playable fragments load. IE: when the duration gets 10 seconds longer, the current time is also 10 seconds longer.

I've traced this as far as I feel comfortable to the following call stack:

1. [when dispatching the timeupdate event](https://github.com/clappr/clappr/blob/0.2.18/src/playbacks/hls/hls.js#L80)
2. [`this.getCurrentTime()` returns the wrong value](https://github.com/clappr/clappr/blob/0.2.18/src/playbacks/hls/hls.js#L45)
3. [which seems to have bad input coming from here](https://github.com/clappr/clappr/blob/0.2.18/src/playbacks/hls/hls.js#L19)
4. [and I don't know why `this.getDuration()` never changes](https://github.com/clappr/clappr/blob/0.2.18/src/playbacks/html5_video/html5_video.js#L200)

I added a `console.log` into `getPlayableStartTime()` to show the weirdness:

screen shot 2015-11-13 at 4 29 09 pm

What you see is `this.getDuration()` (1st number) staying the same while the actual duration (2nd number) steadily increases in 10-second chunks (as expected). This causes `getPlayableStartTime()` to be an ever-growing negative number, in turn causing "current time" as dispatched in the event to grow artificially.

- Any idea what the fix is here?
- Where does `this.el.duration` come from / who is responsible for updating it? Can it be ignored?

tjenkinson:
Also the time which is now emitted in the time update events from the HLS playback is now always the actual time, no rounding occurs. This fixes #668.

As a result new logic has been added to MediaControl so that the seek bar sticks to the right if seek is disabled, or dvr is enabled but not in use.

Tested with a hls stream with sliding window, hls stream with normal dvr (no sliding window), and also a standard mp4 and all seem to work.

### DIFF
--- a/src/components/media_control/media_control.js
+++ b/src/components/media_control/media_control.js
@@ -67,6 +67,8 @@ export default class MediaControl extends UIObject {
     this.persistConfig = this.options.persistConfig
     this.container = options.container
     this.seekTime = new SeekTime(this)
+    this.currentPositionValue = null
+    this.currentDurationValue = null
     var initialVolume = (this.persistConfig) ? Config.restore("volume") : 100
     this.setVolume(this.mute ? 0 : initialVolume)
     this.keepVisible = false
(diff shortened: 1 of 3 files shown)


leandromoreira:
Thanks a lot :+1:

PR #686, opened 2015-11-21
hlsjs: checks if MSE is enabled

leandromoreira:
Why do we need to explicitly skip Safari? (isn't [html5 video](https://github.com/clappr/clappr/blob/master/src/components/loader/loader.js#L47) playback [able to play](https://github.com/clappr/clappr/blob/master/src/playbacks/html5_video/html5_video.js#L267) it?)

@cpeterso is it safe to assume that `!!window.MediaSource` will return `false` for all FF with disabled **MSE**?

I'm trying to test in multiple browsers:

* http://browsershots.org/https://jsfiddle.net/xd93z211/1/

It seems that is reliable.


leandromoreira:
Actually we can even remove this check since [`HLSJS.isSupported()`](https://github.com/dailymotion/hls.js/blob/7584655ebd1de4297c6004ade6b5927f790511ff/src/hls.js#L21) already test that.

Issue #685, opened 2015-11-21
Real time of seekbar


cowai:
Hi,
It should be possible to have the real time in the dvr buffer for the hls stream.

Is that possible today? If not, anyone wanting to point me in the right direction?

Update:
Just to clearify. Instead of showing the dvr buffer duration in the seek bar, show the relevant local time of the seeked position.
That means that "LIVE" will be now, and everything is relative to current time. Makes sense?

cowai:
This issue can be closed as the feature has been implemented.

thiagopnts:
Thanks!

PR #688, opened 2015-11-23
Added actualLiveTime option.

cowai:
screenshot1-crop
Displays the seek time according to current time. (browser time though)
This is meant to be used with a live feed with dvr. Instead of showing seektime relative to the buffer, this makes it relative to the browsers time.
The time of the screenshot was 16:37:25 and the buffer was ~3.5h.
This implements #685.

### DIFF
--- a/src/base/utils.js
+++ b/src/base/utils.js
@@ -23,7 +23,7 @@ export function extend(parent, properties) {
   return MergedPlugin
 }
 
-export function formatTime(time) {
+export function formatTime(time, paddedHours) {
     if (!isFinite(time)) {
       return "--:--"
     }
@@ -40,29 +40,7 @@ export function formatTime(time) {
       out += days + ":"
       if (hours < 1) out += "00:"
     }
-    if (hours && hours > 0) out += ("0" + hours).slice(-2) + ":"
-    out += ("0" + minutes).slice(-2) + ":"
-    out += ("0" + seconds).slice(-2)
-    return out.trim()
-}
-export function formatActualTime(time) {
-    if (!isFinite(time)) {
-      return "--:--"
-    }
-    time = time * 1000
-    time = parseInt(time/1000)
-    var seconds = time % 60
-    time = parseInt(time/60)
-    var minutes = time % 60
-    time = parseInt(time/60)
-    var hours = time % 24
-    var days = parseInt(time/24)
-    var out = ""
-    if (days && days > 0) {
-      out += days + ":"
-      if (hours < 1) out += "00:"
-    }
-    out += ("0" + hours).slice(-2) + ":"
+    if (hours && hours > 0 || paddedHours !== undefined && paddedHours) out += ("0" + hours).slice(-2) + ":"
     out += ("0" + minutes).slice(-2) + ":"
     out += ("0" + seconds).slice(-2)
     return out.trim()
--- a/src/base/utils.js
+++ b/src/base/utils.js
@@ -40,7 +40,7 @@ export function formatTime(time, paddedHours) {
       out += days + ":"
       if (hours < 1) out += "00:"
     }
-    if (hours && hours > 0 || paddedHours !== undefined && paddedHours) out += ("0" + hours).slice(-2) + ":"
+    if (hours && hours > 0 || paddedHours) out += ("0" + hours).slice(-2) + ":"
     out += ("0" + minutes).slice(-2) + ":"
     out += ("0" + seconds).slice(-2)
     return out.trim()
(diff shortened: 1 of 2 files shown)


leandromoreira:
Looks great, I'd also add documentation to [README](https://github.com/clappr/clappr#add-mimetype-for-extension-less-url) and [the code](https://github.com/clappr/clappr/blob/master/src/components/player.js#L70)

src/components/seek_time/seek_time.js
```diff
@@ -99,7 +109,12 @@ export default class SeekTime extends UIObject {
 
       if (this.durationShown) {
         this.$durationEl.show()
-        var currentDuration = formatTime(this.duration)
+        if (this.actualLiveTime) {
+          var currentDuration = formatActualTime(secondsSinceMidnight)
+        } else {
+          var currentDuration = formatActualTime(this.duration)
+        }
```
leandromoreira: how about?
```javascript
var timeToFormat = (this.actualLiveTime? secondsSinceMidnight:this.duration
var currentDuration = formatActualTime(timeToFormat)
```

src/components/seek_time/seek_time.js
```diff
@@ -89,8 +90,17 @@ export default class SeekTime extends UIObject {
...
+      if (this.actualLiveTime) {
+        var d = new Date(), e = new Date(d);
+        var secondsSinceMidnight = (e - d.setHours(0,0,0,0)) / 1000;
+        var seekTime = (secondsSinceMidnight - this.duration) + (this.hoverPosition * this.duration)
+        if (seekTime < 0) {
+          seekTime = (3600 * 24) + seekTime
+        }
+      } else {
+        var seekTime = this.hoverPosition * this.duration
+      }
```
leandromoreira: maybe we could extract this to a method

cowai: I totally agree on all your points. If you guys are okay with inline if statements I agree it will actually be a little cleaner. For the redundant formatActualTime(), I agree we could check for the actualLiveTime option inside formatTime() as there were only a little line change to get zero-filled hours. I will push some code later :)

cowai: I came to think of a possible more flexible option. How about allowing a date object as an option where we add seek position/duration to that date. This way we can set server side time and let users go around the problem with wrong timezones and other client side problems.

src/base/utils.js
```diff
@@ -40,7 +40,7 @@ export function formatTime(time) {
       out += days + ":"
       if (hours < 1) out += "00:"
     }
-    if (hours && hours > 0) out += ("0" + hours).slice(-2) + ":"
+    if (hours && hours > 0 || paddedHours !== undefined && paddedHours) out += ("0" + hours).slice(-2) + ":"
```
tjenkinson: Think this can be simplified to just
```js
if (hours && hours > 0 || paddedHours)
```

src/components/seek_time/seek_time.js
```diff
@@ -30,6 +30,7 @@ export default class SeekTime extends UIObject {
     this.hoveringOverSeekBar = false
     this.hoverPosition = null
     this.duration = null
+    this.actualLiveTime = this.mediaControl.options.actualLiveTime === true
```
tjenkinson: This can just be:
```js
this.actualLiveTime = this.mediaControl.options.actualLiveTime
```

src/components/seek_time/seek_time.js
tjenkinson: I like the idea of the server providing the stream start time but don't think that will work with streams with a sliding window where the start of the stream is removed as time progresses.

src/base/utils.js
cowai: I wasn't sure if formatTime was used elsewhere, if so paddedHours wouldn't be a boolean.

tjenkinson: In javascript `null`, `undefined` and `false` are all falsey so if it's `undefined` it's still ok.

cowai: Good to know :)  Does this work in all common browsers too?

src/components/seek_time/seek_time.js
cowai: What I meant was providing server time to what the stream is now at its freshest point, and then calculate what the beginning of the stream is by calculating how long since initiation and seek time. That would work I think.

src/base/utils.js
tjenkinson: Yep :)
http://james.padolsey.com/javascript/truthy-falsey/

cowai:
https://github.com/clappr/clappr/pull/693
Second attempt at this.

I am looking forward to hearing your thoughts. Especially about servertime and using a string to parse.

Issue #533, opened 2015-08-26
Bundle JS seems to fail on the new release


lkkadiri:
Our bundleJS task is failing with 0.2.7, and throws the following error

```
/custom-player/node_modules/clappr/src/main.js:5
import Player from './components/player'
^
ParseError: 'import' and 'export' may appear only with 'sourceType: module'
```

Any help with this will be great.

lkkadiri:
@jfairley @dviramontes

jfairley:
because of this https://github.com/clappr/clappr/commit/b309deb795fa0f02cd024f46751385972b1dd3fd

towerz:
@lkkadiri @jfairley it seem like the problem is the entry point for clappr is affecting your build configuration. I'm investigating whether there's a solution for this or if I'll need to revert this change.

What kind of build configuration are you using? I mean, how's clappr included in your build?

dviramontes:
@towerz, 
we are using **browserify**.

dviramontes:
As a temporary fix you can add a browser property to your package.json
(if you are using **browserify**):
```
  "browser": {
    "clappr": "./node_modules/clappr/dist/clappr.min.js"
  }
```

lkkadiri:
Making the change suggested by @dviramontes fixes the build but now the min version of clappr does not seem to find the swf file. I now see this error,

```
Failed to load resource: the server responded with a status of 404 (Not Found)
http://localhost:3000/js/96f944f0104ee30b8fce6cffd89e13aa.swf?inline=1 
```

towerz:
Just reverted b309deb795fa0f02cd024f46751385972b1dd3fd to avoid breaking browserify usages. I found a way to expose source/assets to webpack without breaking other build systems.

Will release 0.2.8 asap.

towerz:
Done, closing the issue now. If you have further problems, please let me know.

lkkadiri:
@towerz I still get this error, 

```
GET http://localhost:3000/js/96f944f0104ee30b8fce6cffd89e13aa.swf?inline=1  404 (Not Found)
```
Seems like its unable to find the swf file still.

towerz:
@lkkadiri prior to this change, where were clappr getting the swf from?

[...]

PR #690, opened 2015-11-24
Build now works on Travis Container Infrastructure

tjenkinson:
Tested a build locally and it still appears to work but this will need testing with different formats to make sure I haven't removed something important.

Chrome and Firefox browsers now work and are used for running the tests. I have fixed one of the tests which was failing in Firefox.

Removed jquery dependency from tests and clappr-zepto now used and required explicitly in the place it was required.

Updated some of the versions of some of the dependencies.

Removed "browser" property from `package.json`. Not sure what it was there for, or if it was important?

Moved dependencies from "dependencies" to "dev-dependencies" because I don't think any of the ones that were there were required for using Clappr, because everything that is needed is bundled into the `clappr.js` file. This means `npm-shrinkwrap.json` is empty as this by default only locks the `dependencies` list but I think this still makes sense.


leandromoreira:
Great work @tjenkinson =) congrats, I'll try to test this build later (with hls, mp4 ... )

leandromoreira:
> My understanding is that dev dependencies are the dependencies needed for a developer to work on the project and built it etc. `dependencies` are the dependencies that are required in order for the project to be run/used by a user. 

I think the same.

> Because the entire project is bundled into the clappr.min.js file this is all the user needs in order to use the project. The entry point that the end user should use to run the project is what's specified by the main property, which is dist/clappr.min.js.

Yes too.

> For example if someone else in a separate project does `npm install clappr` they will get clappr.min.js and that's it.

Yes too but I think if you move those dependencies to `devDependencies`, the final bundle **clappr.min.js** will throw an error because `uniq` and the others are not contained within it. 

That's my understanding (I grasp that from using: bundler, pip, maven, gradle ...) but I really don't have 100% confidence with `npm`. What you can do to test is:

```shell
vim package.json #(move everything to devDependency)
rm -rf node_modules && rm npm-shrinkwrap.json
npm install
npm start
```
And try to see on browser if there is any error.

leandromoreira:
You're right (that's crazy), as I said of all those package managers (bundler, maven, ...) all of them exclude libs not declared as `dependency`. In the end webpack will do the job of building clappr.min.js

But anyway, I still think we need to **document** the intention under `dependency` because declaring it as [`devDependency`](http://stackoverflow.com/a/19341028) we're assuming that they are not required for production.

>**devDependencies** are for the development-related scripts, e.g. unit testing, packaging scripts, documentation generation, etc.

>**dependencies** are required for production use, and assumed required for dev as well.

towerz:
@tjenkinson actually you are correct in your understanding. Clappr declares the main module as clappr.min.js in package.json (which makes unnecessary to declare its dependencies), but its source code is included in the npm package as well. This allows a user to include Clappr in a project codebase, generating a single bundle.

So in order to use Clappr by importing its source, it would be necessary to redeclare its own dependencies in the target project. Using `devDependencies` doesn't fit this well, because the user won't be able to differentiate what is required for building Clappr from what's only a development tool used in the Clappr project. In this case, using `dependencies` instead of `devDependencies` serves this purpose well, like @leandromoreira mentioned.

In fact, we only declare clappr.min.js as entry point in order to allow projects which do not use ES6 to include Clappr as dependency without needing to compile it.

towerz:
@tjenkinson we had already done that before, pointing main to src/main.js, but there were some complaints after we migrated to webpack. There were people who used Clappr with browserify that weren't able to build their projects without including babel and other dev dependencies, because they couldn't set their projects to use clappr.min.js easily, at least not without breaking asset references (like swf paths, etc). This was described in #533.

Since changing this back to clappr.min.js didn't affect us, we gave in and changed the main back to the build output dist/clappr.min.js. We can check whether these problems do still occur and change main back to src/main.js.

towerz:
Yes, it seems to be the case. And to make things worse, with browserify the user has to manually manage the player assets when compiling the bundle, making it very difficult to avoid errors.

towerz:
I agree it's not complicated to switch from browserify to webpack, but I don't want to force this change on anyone. About your first point, if `browser` is added, it doesn't mean that will work for browserify users (it will still present the assets issue), but we can add instructions for browserify users to do that.

If you want to proceed with that, I'd like to ask you to revert the devDependency changes and switch the main reference to src/main.js, and add instructions to make browserify work correctly with assets in the readme file.

towerz:
I'm not sure, theoretically it shouldn't make a difference, but since people have had problems with this change before, perhaps we need to provide clearer instructions for using Clappr with browserify. I just asked you to do so because you are already changing the package spec, and I think we need to better document these changes during the process.

towerz:
I wouldn't recommend to use the built version when using webpack, unless it's declared as `external`. It would introduce some unnecessary overhead to the built product.

In webpack, to override the package main when building, we use the following alias in webpack.config.js in our custom player:
```javascript
resolve: {
    alias: { Clappr: 'clappr/src/main.js' },
    root: [path.resolve(__dirname, 'node_modules/clappr/src')],
    extensions: ['', '.js'],
}
```
This makes webpack use src/main.js as Clappr package entry point (when requiring/importing 'Clappr') and use clappr/src as a resolve path (in order to resolve 'plugins/poster' for example).

.travis.yml
```diff
@@ -1,24 +1,13 @@
...
-  - "sh -e /etc/init.d/xvfb start"
-
-install:
-  - "sudo apt-get install -qq gcc-4.8 g++-4.8"
-  - "sudo update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-4.8 90"
-  - "npm install"
+  - "4.2.2"
 
-env:
-  - CHROME_BIN=/usr/bin/chromium-browser COVERALLS_SERVICE_NAME=travis-ci COVERALLS_REPO_TOKEN=sj8SOxUBGaVeFGSY9ryrC2eTFpVjCPHJl
```
leandromoreira: I think it's still good to send coveralls, but we can do that after the merge :+1:

tjenkinson: It's back now :) wasn't sure if it was doing anything before

PR #693, opened 2015-11-26
Introducing actualLiveTime feature

cowai:
New option actualLiveTime gives us an actual time in the seek bar.

Along with actualLiveTime there is an option, servertime which can be used to override the local time used with actualLiveTime
servertime is specified right now as a string like "2015/11/26 06:01:03". Since the code uses the Date object in javascript, wrong timezones and wrong time locally is a problem, but
actually parsing the servertime as localtime makes this go away.

I deleted my first repo so I needed to make a second pull request.
Old one for reference: https://github.com/clappr/clappr/pull/688


leandromoreira:
the `servertime` feature is pretty neat, congrats!

leandromoreira:
I'd rename `serverTime` to `actualLiveServerTime`.

leandromoreira:
I was wondering how about you make this an external plugin? it'd be a great plugin ;)

cowai:
@leandromoreira To do that would I need to essentially override the seek_time component? 
I would guess everybody that is using Clappr for live feeds would be wanting this though as it not very user friendly to display a random seek time which increases over time.

leandromoreira:
@cowai :) thanks, you're it'd probably require to override it... I'll talk with some members soon (which are working on redoing media control/seek time)

cowai:
Thanks :)

src/components/seek_time/seek_time.js
```diff
@@ -30,6 +30,15 @@ export default class SeekTime extends UIObject {
...
     this.duration = null
+    this.actualLiveTime = !!this.mediaControl.options.actualLiveTime
+    if (this.actualLiveTime) {
+      if (!!this.mediaControl.options.actualLiveServerTime) {
+        this.actualLiveServerTimeDiff = new Date().getTime() - new Date(this.mediaControl.options.actualLiveServerTime).getTime()
+      } else {
+        this.actualLiveServerTimeDiff = 0
+      }
+    }
+    console.log("actualLiveServerTimeDiff: " + formatTime(this.actualLiveServerTimeDiff / 1000))
```
leandromoreira: :x: console.log

cowai: Fixed!

leandromoreira:
@cowai thank you so much and sorry for making you wait so long.

cowai:
Awesome. Thanks!

leandromoreira:
man your idea was cool! :)

cowai:
Ehm, was this directed at me?

leandromoreira:
yep @cowai we're having insights about its usage.

cowai:
Ah thanks :) Care to elaborate on your insights? :)

leandromoreira:
We do have lots of live streams, some of them will be much much easier to use (dvr) if we provide the current time and then we can build a UX over it.

cowai:
@leandromoreira @tjenkinson Nice to hear this is of value to you :)
For my use case, it makes sense as the viewers would want to follow certain programs, and being able to skip to a certain point based on current time is much more user friendly.

cowai:
Providing server time is also important as in my experience people have crazy setupts with both wrong time zones setup for their location and wrong time to compensate for that. Along with time drift, it means we cannot trust the users local time in any way. It also makes it possible to have a UX for setting the timezone for which the time should be initalized.

cowai:
Have you tried passing it? It should work. It will create a new date object though.

PR #698, opened 2015-11-27
Refactored seek method  to accept values in seconds instead of percentage (fixes #654)

### LINKED ISSUE #654: player.seek() should accept seconds not percent
This command should seek to the current video position
```js
player.seek(player.getCurrentTime());
```
In documentation:
##  player.seek(value);
The `value` should be a number between 0 and 100. For example, `player.seek(50)` will seek to the middle of the current source.

dayvson:


src/components/container/container.js
```diff
@@ -251,9 +251,13 @@ export default class Container extends UIObject {
     this.trigger(Events.CONTAINER_CONTEXTMENU, this, this.name)
   }
 
-  setCurrentTime(time) {
-    this.trigger(Events.CONTAINER_SEEK, time, this.name)
-    this.playback.seek(time)
+  setCurrentTime(time, inseconds) {
```
leandromoreira: maybe we can assume `false` as default (so we won't broken anything) but I also think that the default should be **seconds** =/
```javascript
setCurrentTime(time, inseconds=false)
```

leandromoreira:
awesome :) looks GREAT to me

src/components/container/container.js
```diff
@@ -251,9 +251,13 @@ export default class Container extends UIObject {
...
+    this.trigger(Events.CONTAINER_SEEK, time, this.name);
+    if(inseconds) {
```
flavioribeiro: missing space `if (inseconds) {`

src/components/container/container.js
```diff
@@ -251,9 +251,13 @@ export default class Container extends UIObject {
...
+      this.playback.seek(time);
+    }else{
```
flavioribeiro: missing space `} else {`

src/playbacks/flash/flash.js
```diff
@@ -177,24 +177,24 @@ export default class Flash extends BaseFlashPlayback {
     return this.el.getDuration()
   }
 
-  seek(seekBarValue) {
+  seekPercentage(seekBarValue) {
     if (this.el.getDuration() > 0) {
       var seekTo = this.el.getDuration() * (seekBarValue / 100)
```
flavioribeiro: maybe change this var from `seekTo` to `seekSeconds`?

src/playbacks/flashls/flashls.js
```diff
@@ -412,12 +412,16 @@ export default class FlasHLS extends BaseFlashPlayback {
     return duration
   }
 
-  seek(time) {
+  seekPercentage(percentage) {
     var duration = this.el.getDuration()
-    if (time > 0) {
-      time = duration * time / 100
+    if (percentage > 0 ) {
```
flavioribeiro: extra space `if (percentage > 0) {`

src/components/container/container.js
flavioribeiro: I'm in favor of not having this flag, we can create another method to be called by `seekPercentage()`

src/components/player.js
```diff
@@ -250,13 +250,23 @@ export default class Player extends BaseObject {
...
+
   /**
    * seeks the current video (`source`). For example, `player.seek(50)` will seek to the middle of the current video.
    * @method seek
    * @param {Number} time should be a number between 0 and 100.
    */
-  seek(time) {
-    this.core.mediaControl.container.setCurrentTime(time);
+  seekPercentage(percentage) {
+    this.core.mediaControl.container.setCurrentTime(percentage);
```
flavioribeiro: here, instead of calling `setCurrentTime()` we can create a new method on the container to avoid the flag and the `if`

src/components/container/container.js
leandromoreira: :+1:

dayvson: I already created another method called seekPercentage. 
In my PR, there are two public methods available seek(seconds) seekPercentage(percentage) and that setCurrentTime is a private method.

src/components/player.js
dayvson: I agree with you, but I don't think this is that critical, we would have two methods that are almost the same both needs to dispatch the CONTAINER_EVENT so instead of this: 
```
setCurrentTime(time, inseconds) {
    this.trigger(Events.CONTAINER_SEEK, time, this.name);
    if (inseconds) {
      this.playback.seek(time);
    } else {
      this.playback.seekPercentage(time);
    }
  }
```
So we would have something like this:
```
setCurrentTime(time) {
    this.trigger(Events.CONTAINER_SEEK, time, this.name);
    this.playback.seek(time);
  }

setCurrentPercentage(percentage) {
    this.trigger(Events.CONTAINER_SEEK, percentage, this.name);
    this.playback.seekPercentage(percentage);
}
```

leandromoreira: Since we're going to dispatch two kinds of values on `CONTAINER_SEEK` event, maybe we need to redesign it:
```javascript
this.trigger(Events.CONTAINER_SEEK, {percentage: true, value: percentage}, this.name)
this.trigger(Events.CONTAINER_SEEK, {percentage: false, value: time}, this.name)
//or
this.trigger(Events.CONTAINER_SEEK_PERCENTAGE, percentage, this.name)
this.trigger(Events.CONTAINER_SEEK, time, this.name)
```

I'm in favor of the second (adding another event)

src/components/container/container.js
dayvson: This will break the API and people using the latest need to update their code. 
But it only will break for people using the API if they are using just the player with the regular controls it will work just fine.

src/components/player.js
dayvson: I broke down the setCurrentTime into two methods seek and seekPercentage, and it looks much better. Also, the CONTAINER_SEEK event will be always triggered with seconds.

src/components/player.js
```diff
@@ -250,13 +250,23 @@ export default class Player extends BaseObject {
...
+   * @method seekSeconds
+   * @param {Number} time should be a number between 0 and the video duration.
+   */
+  seek(time) {
+    this.core.mediaControl.container.seek(time);
+  }
+
   /**
    * seeks the current video (`source`). For example, `player.seek(50)` will seek to the middle of the current video.
    * @method seek
```
leandromoreira: missing change `seek` to `seekPercentage`

src/base/playback.js
```diff
@@ -37,13 +37,20 @@ export default class Playback extends UIObject {
...
+   * seeks the playback to a given `time` in seconds
    * @method seek
-   * @param {Number} time should be a number between 0 and 100
+   * @param {Number} time should be a number between 0 and the video duration
    */
   seek(time) {}
 
   /**
+   * seeks the playback to a given `percentage` in percentage
+   * @method seek
```
leandromoreira: `seek` to `seekPercentage`

src/components/player.js
```diff
@@ -250,13 +250,23 @@ export default class Player extends BaseObject {
     this.core.mediaControl.container.stop();
   }
 
+  
+  /**
+   * seeks the current video (`source`). For example, `player.seek(120)` will seek to second 120 (2minutes) of the current video.
+   * @method seekSeconds
```
leandromoreira: `seekSeconds` to `seek`

src/playbacks/flashls/flashls.js
```diff
@@ -412,12 +412,16 @@ export default class FlasHLS extends BaseFlashPlayback {
     return duration
   }
 
-  seek(time) {
+  seekPercentage(percentage) {
     var duration = this.el.getDuration()
-    if (time > 0) {
-      time = duration * time / 100
+    if (percentage > 0) {
+      var time = percentage * 60 / duration
```
leandromoreira: maybe we can:
* start it with zero, outside of  ifs `var time = 0` in order to avoid `seek(undefined)`
* or put a check `!!time && this.seek(time)`

I'm in favor of the last

src/playbacks/hls/hls.js
```diff
@@ -54,15 +54,19 @@ export default class HLS extends HTML5VideoPlayback {
...
-    if (seekBarValue > 0) {
-      seekTo = this.playableRegionDuration * (seekBarValue / 100)
+    if (percentage > 0) {
+      seekTo = this.playableRegionDuration * (percentage / 100)
     }
-    var onDvr = this.dvrEnabled && seekBarValue >= 0 && seekBarValue < 100
-    seekTo += this.playableRegionStartTime
-    super.seekSeconds(seekTo)
-    this.updateDvr(onDvr)
+    this.seek(seekTo)
```
leandromoreira: :+1: cool

src/playbacks/flashls/flashls.js
tjenkinson: With the second one won't that mean `this.seek(time)` will never be called when `time` is 0?

leandromoreira: You're right! maybe this one could work:
`(time !== undefined) && this.seek(time)`

flavioribeiro:
nice, good job @dayvson :beers:

Issue #707, opened 2015-12-03
Expose flashls playback configuration parameters as embed options


me-vlad:
Hello,

I want to discuss this question before submitting pull requests to clappr and clappr/flashls.
Flashls has many configuration parameters defined in https://github.com/mangui/flashls/blob/dev/src/org/mangui/hls/HLSSettings.as

What do you think about adding these settings (expose) to clappr flashls component (with new overridden functions in https://github.com/clappr/flashls/blob/clappr/src/io/clappr/Player.as)?

```
    override protected function _setupExternalCallers() : void {
      super._setupExternalCallers();
      ExternalInterface.addCallback("playerSetAutoStartLoad", _setAutoStartLoad);
      ExternalInterface.addCallback("playerSetCapLevelToStage", _setCapLevelToStage);
      ExternalInterface.addCallback("playerSetMaxLevelCappingMode", _setMaxLevelCappingMode);
      ExternalInterface.addCallback("playerSetMinBufferLength", _setMinBufferLength);
      ExternalInterface.addCallback("playerSetMinBufferLengthCapping", _setMinBufferLengthCapping);
      ExternalInterface.addCallback("playerSetMaxBufferLength", _setMaxBufferLength);
      ExternalInterface.addCallback("playerSetMaxBackBufferLength", _setMaxBackBufferLength);
      ExternalInterface.addCallback("playerSetLowBufferLength", _setLowBufferLength);
      ExternalInterface.addCallback("playerSetMediaTimePeriod", _setMediaTimePeriod);
      ExternalInterface.addCallback("playerSetFpsDroppedMonitoringPeriod", _setFpsDroppedMonitoringPeriod);
      ExternalInterface.addCallback("playerSetFpsDroppedMonitoringThreshold", _setFpsDroppedMonitoringThreshold);
      ExternalInterface.addCallback("playerSetCapLevelonFPSDrop", _setCapLevelonFPSDrop);
      ExternalInterface.addCallback("playerSetSmoothAutoSwitchonFPSDrop", _setSmoothAutoSwitchonFPSDrop);
      ExternalInterface.addCallback("playerSetSwitchDownOnLevelError", _setSwitchDownOnLevelError);
      ExternalInterface.addCallback("playerSetSeekMode", _setSeekMode);
      ExternalInterface.addCallback("playerSetKeyLoadMaxRetry", _setKeyLoadMaxRetry);
[...]

PR #712, opened 2015-12-04
Add flashls playback configuration parameters

me-vlad:
Expose flashls playback configuration parameters as embed options #707


towerz:
Thanks @me-vlad!

PR #716, opened 2015-12-06
Make mime type take precedence over extension if provided in HTML5AudioPlayback

tjenkinson:


leandromoreira:
I think this is very similar to what html5video does. We can test it to see and if it's possible I'd suggest to reuse [`html5video.canplay`](https://github.com/clappr/clappr/blob/master/src/playbacks/html5_video/html5_video.js#L293) in order to avoid duplication:

> if you look closely both `returns` ([1](https://github.com/clappr/clappr/blob/master/src/playbacks/html5_audio/html5_audio.js#L45) and [2](https://github.com/clappr/clappr/blob/master/src/playbacks/html5_audio/html5_audio.js#L48)) relies on `canPlayType` :P

```javascript
//introduce the common pattern
HTML5Video._canPlay(type, mimetypes, resource, mimeType) {
 var extension = (srcUrl.split('?')[0].match(/.*\.(.*)$/) || [])[1]
 var mimeTypes = mimetypes[extension] || mimeType
  mimeTypes = (mimeTypes.constructor === Array) ? mimeTypes : [mimeTypes]

  var media = document.createElement(type)
  return !!find(mimeTypes, (mediaType) => !!media.canPlayType(mediaType).replace(/no/, ''))
}
//use it on video
HTML5Video.canPlay(resource, mimeType) {
  return HTML5Video._canPlay('video', MIMETYPES, resource, mimeType)
}
//use it on audio
HTML5Audio.canPlay(resource, mimeType) {
 var mimetypes = {
   'wav': ['audio/wav'],
   'mp3': ['audio/mp3', 'audio/mpeg;codecs="mp3"'],
   'aac': ['audio/mp4;codecs="mp4a.40.5"'],
   'oga': ['audio/ogg']
 }
  return HTML5Video._canPlay('audio', mimetypes, resource, mimeType)
}

```

What do you think?

leandromoreira:
cool @tjenkinson I'd suggest to update the documentation on [playback.canPlay](https://github.com/clappr/clappr/blob/master/src/base/playback.js#L122) to reflect this direction ** if the user is providing the mime type that they know the content is, then that should always be used.**

I'm sorry for being so [cranky](http://vignette2.wikia.nocookie.net/donkeykong/images/9/96/Cranky_Complaining.gif/revision/latest?cb=20100815235935)

leandromoreira:
LGTM I'll merge, as always thanks @tjenkinson

PR #717, opened 2015-12-06
Fix bug which would sometimes prevent video playing when overriding mimetype

tjenkinson:
Previously if the user was specifying a mime type, it was not carried through when the check was performed again in `setupSrc()`


leandromoreira:
@tjenkinson since this `setupSrc` is new the only playback that would need to override would be dash :P I think this PR it's pretty cool :+1:

leandromoreira:
=) I think it's good to remove that on this PR

PR #722, opened 2015-12-08
flashls: update HLSPlayer.swf to 0.4.9 (flashls 0.4.4.16) and flashlsChromeless.swf to 0.4.4.16

me-vlad:


towerz:
Thanks @me-vlad!

PR #723, opened 2015-12-08
Fix so that volume level doesn't get replaced by MediaControl

tjenkinson:
Previously if the user did `player.setVolume(something)` immediately after loading the player, it would get reset on the next tick due to a media control `render()`.


towerz:
Merged, thanks @tjenkinson!

PR #730, opened 2015-12-11
Added ready callback argument

tjenkinson:


src/components/player.js
```diff
@@ -99,8 +99,11 @@ export default class Player extends BaseObject {
    * define a poster by adding its address `poster: 'http://url/img.png'`. It will appear after video embed, disappear on play and go back when user stops the video.
    * @param {String} [options.playbackNotSupportedMessage]
    * define a custom message to be displayed when a playback is not supported.
+   * @param {Object} readyCallback (optional)
+   * If provided this will be called when the player has finished loading (just before the READY event is fired).
+   * It will be provided the player instance as the first argument.
    */
-  constructor(options) {
+  constructor(options, readyCallback) {
```
leandromoreira: Why wouldn't you just do this:

```javascript
//new player
player.on(Clappr.Events.PLAYER_READY, function(){})
```

tjenkinson: I could, just thought it would be nicer to have it available as an argument so the user doesn't have to register for the event as well as creating the player.

If you'd rather people just listen for the event though that's fine just close this.

leandromoreira: We can discuss that with @clappr/core :) with that said I think it'd be nicer if we provide it as an option, then we can extend to a set of most used events (onPause, onPlay, onStop, onError(this one might be pretty useful) ...):

```javascript
new Clappr.Player({
events: {
  onReady: "..",
  onPlay: "..."
}
})
```

tjenkinson: Just discovered an issue with events which is that they can be fired before the `new Clappr.Player()` completes, meaning the user doesn't get chance to listen to them!

@leandromoreira's solution would allow people to listen to events before the creation of the player completes, if the listeners for them were registered internally before anything would fire them.

towerz:
I agree with @leandromoreira, this would be better implemented in the `options`. The listeners could be registered internally when creating the core/containers, so they would be registered before events are fired. What do you think, @tjenkinson?

leandromoreira:
@tjenkinson I think this is the initial list:

* onReady
* onResize
* onPlay
* onPause
* onStop
* onEnded
* onSeek
* onError
* onTimeUpdate
* onVolumeUpdate

PR #732, opened 2015-12-11
Added core.isReady(), player.isReady(), and make sure PLAYER_READY not missed

tjenkinson:
It's possible for the `CORE_READY` event to be fired before the `create()` method returns, making it impossible to listen to the event.

Therefore added `isReady()` method, and this is used to determine if event has been missed.


towerz:
Great addition, @tjenkinson!

Issue #731, opened 2015-12-11
Support multiple sources with different user-provided mime types


tjenkinson:
I think at the moment this isn't possible.

At the moment you can provide a mimetype with the `mimeType` option, but this then applies to all `sources`. I need to be able to provide multiple sources, which have different mimetypes.

Proposed solution:

```javascript
new Clappr.Player({
   sources: [
      {source: "video.mp4", mimeType: "video/mp4"},
      {source: "video.m3u8", mimeType: "application/x-mpegURL"}
   ]
});
```

For backwards compatibility, if an item in the array is a string, the behaviour will stay the same as it is now.

If `mimeType` is also provided, it will be used like now, but `mimeType`'s in source objects would take precedence.

Sound ok?

tjenkinson:
This would also apply to the `sources` parameter of the `load(sources, mimeType)` method.

towerz:
Closing the issue, #733 is already merged.

Allan-Nava:
> I think at the moment this isn't possible.
> 
> At the moment you can provide a mimetype with the `mimeType` option, but this then applies to all `sources`. I need to be able to provide multiple sources, which have different mimetypes.
> 
> Proposed solution:
> 
> ```js
> new Clappr.Player({
>    sources: [
>       {source: "video.mp4", mimeType: "video/mp4"},
>       {source: "video.m3u8", mimeType: "application/x-mpegURL"}
>    ]
> });
> ```
> 
> For backwards compatibility, if an item in the array is a string, the behaviour will stay the same as it is now.
> 
> If `mimeType` is also provided, it will be used like now, but `mimeType`'s in source objects would take precedence.
> 
> Sound ok?

this config sources it works?

kslimani:
Yes it "works" (ie: it load failover source if previous failed to be resolved). But it seems there an issue with mediacontrols (not displayed) when failover source is loaded. _(events seems to not be properly binded)_

Allan-Nava:
> Yes it "works" (ie: it load failover source if previous failed to be resolved). But it seems there an issue with mediacontrols (not displayed) when failover source is loaded. _(events seems to not be properly binded)_

thanks for the support, i try this example of configuration

PR #733, opened 2015-12-11
Added support for mime types per source

tjenkinson:
Just put together what I proposed in #731 

If you think there is a better way or want it tweaking that's fine.


towerz:
I think this is a great feature, thanks @tjenkinson!

Issue #737, opened 2015-12-15
Test fail for current HEAD (#052f9b9a5)


avreg:
```
Chrome 47.0.2526 (Linux 0.0.0) Player constructor should trigger error events FAILED
	AssertionError: expected spy to have been called at least once, but it was never called
	    at Context.<anonymous> (/var/tmp/clappr/test/player_spec.js:114:25)
Iceweasel 38.4.0 (Linux 0.0.0) Player constructor should trigger error events FAILED
	expected spy to have been called at least once, but it was never called
	AssertionError@/var/tmp/clappr/node_modules/chai/chai.js:5284:18
	[3]</module.exports/Assertion.prototype.assert@/var/tmp/clappr/node_modules/chai/chai.js:206:1
	sinonProperty/<@/var/tmp/clappr/node_modules/sinon-chai/lib/sinon-chai.js:73:13
	addProperty@/var/tmp/clappr/node_modules/chai/chai.js:3992:22
	@/var/tmp/clappr/test/player_spec.js:114:10
```

tjenkinson:
Just been looking at this and I think the route of the issue is that the playback currently fires a `PLAYBACK_READY` event **or** a `PLAYBACK_ERROR` event but never both.

Code in the core expects the playback always to be ready at some point in order to continue, and then the player can be classed as ready. Therefore if there is an error the player never completes initialisation, and the event listener has never been registered to throw the `PLAYER_ERROR` event.

So the decision needs making as to whether the ready events should always be fired if there's an error or not, and if they shouldn't be then I guess the core should listen for the error events as well and continue on those. @clappr/core ?

towerz:
@tjenkinson the ready event should be fired whenever the player is initalized for playback, either from a initializer/constructor call or a load. When loading a new source, the player behaves like it's being reinitialized, and in fact it is.

towerz:
The test failed due to a change I introduced. I'm still checking whether there's an error in this test assertions or if there's something else to be changed.

tjenkinson:
@towerz ok I understood it differently then I'll close that PR.

I think it's because the container event listeners are now registered when the player is ready, but the player never becomes ready, because only a `PLAYBACK_READY` will allow this to happen.

towerz:
@avreg please refrain from opening issues for broken tests. It's not really a bug or feature request, it's a consequence of the product development.

[...]

PR #739, opened 2015-12-15
Сustom and reorder a set of internal plugins

avreg:
Example (from public/plugins.html):

```
    // override or/and reorder internal plugins set for all players
    Clappr.defaultPlugins.playback = ['flash', 'html5_video', 'html5_audio', 'hls', 'flashls', 'html_img' /*, 'no_op' */];
    Clappr.defaultPlugins.container = ['spinner', 'watermark', 'poster', /* 'stats', 'google_analytics', */ 'click_to_pause'];
    Clappr.defaultPlugins.core = ['dvr_controls', 'favicon', 'seek_time'];

    // override or/and reorder internal plugins set for a current player
        mute: true,
        height: 240,
        width: 320,
        internalPlugins: {
          playback: ['html5_video', 'html5_audio', 'flash', 'flashls', 'hls', 'html_img' /*, 'no_op' */],
          container: ['spinner']
        },
        plugins: {'playback': [RTMP]},
        rtmpConfig: {
          swfPath: 'dist/assets/RTMP.swf',
          scaling:'stretch',
          playbackType: 'live',
          bufferTime: 1,
          startLevel: 0,
          autoSwitch: true,
        }
      });
```


towerz:
@avreg thanks for suggesting this, but it's not really necessary to add overrides for internal plugins. In order to change plugin priority, like playbacks, you could just add them in the order you want in the `plugins` parameter.

When merging the plugin lists, the ones in the `plugins` parameters are added first, and internal plugins with the same name are discarded. This way, if you want to create a new spinner or poster plugin, you could just add plugins with the same name as their internal counterparts to the list.

I think this may be undocumented, so it's a good thing you pointed out that we are missing this explanation in our documentation.

There's one thing not covered by plugin overrides, removing/disabling plugins at instantiation. We could implement this as a plugin or in the core loader, but it's different from specifying the plugin bundle.

In the coming versions we will remove most internal plugins and release them in their standalone form, so you can create your own bundle. We'll add a way to bundle plugins more easily as well.

towerz:
Just to make things clearer, we don't want to expose internal elements of the player in the API. Things that should be public are either part of the API or plugins themselves.

As you mentioned in #737, you need the ability to remove some plugins from the list. Since we don't have this implemented yet, why don't you try to disable the undesired plugins for now? Another way to do this is extending the player class and making your own loader exposing your list of internal plugins.

If you need help doing this I can provide you some examples.

avreg:
В Tue, 15 Dec 2015 08:05:33 -0800
Bruno Torres <notifications@github.com> пишет:

> @avreg thanks for suggesting this, but it's not really necessary to add overrides for internal plugins. In order to change plugin priority, like playbacks, you could just add them in the order you want in the `plugins` parameter.
> 
> When merging the plugin lists, the ones in the `plugins` parameters are added first, and internal plugins with the same name are discarded. This way, if you want to create a new spinner or poster plugin, you could just add plugins with the same name as their internal counterparts to the list.

1) How should look like a global reordering if the page a lot of players?

var override_playbacks = [class1, class2, ...]

player1 = new Clappr.Player({
	plugins: {
		playback: override_playbacks
	},
}
player2 = new Clappr.Player({
	plugins: {
		playback: override_playbacks
	},
}

It seems to me that the visibility like this:

Clappr.defaultPlugins.playback  = [...]

player2 = new Clappr.Player({...})



2) For example, I would like to remove the "hls" plugin for some browsers.

The only way to do it is now:
if (browser is ... and browser_version ... ) {
	override_playbacks = [
                    HTML5VideoPlayback,
		    HTML5AudioPlayback,
                    FlashVideoPlayback,
                    FlasHLSVideoPlayback,
                    HTMLImgPlayback,
                    NoOp
		]
                // !!! trailing NoOp is necessarily
} else {
	override_playbacks = []
}
player1 = new Clappr.Player({
	plugins: {
		playback: override_playbacks
	},
}

Yes, this will work, but is it comfortable?

3) Use plugin class names in the "plugins" parameter may not be a good idea.
For example:
  * SpinnerThreeBouncePlugin vs "spinner",
  * WaterMarkPlugin vs "watermark".

 
> I think this may be undocumented, so it's a good thing you pointed out that we are missing this explanation in our documentation.
> 
> There's one thing not covered by plugin overrides, removing/disabling plugins at instantiation. We could implement this as a plugin or in the core loader, but it's different from specifying the plugin bundle.
> 
> In the coming versions we will remove most internal plugins and release them in their standalone form, so you can create your own bundle. We'll add a way to bundle plugins more easily as well.

Yes, some internal (core,container) plugins are now completely "external".
Their moving from internal solve the problem.
But "playback" plugins are likely to remain as internal,
and again we get only what I wrote above.

avreg:
В Tue, 15 Dec 2015 08:13:18 -0800
Bruno Torres <notifications@github.com> пишет:

> As you mentioned in #737, you need the ability to remove some plugins from the list. Since we don't have this implemented yet, why don't you try to disable the undesired plugins for now?

Really? Without patching loader.js?

How can I just turn off the following plugins:
  * container: [WaterMarkPlugins, StatsPlugin, GoogleAnalyticsPlugin]
  * playback: [hls, flash]
?

> Another way to do this is extending the player class and making your own loader exposing your list of internal plugins.
Yes, it is possible, but more like a hack.

avreg:
В Tue, 15 Dec 2015 08:13:18 -0800
Bruno Torres <notifications@github.com> пишет:

> Another way to do this is extending the player class and making your own loader exposing your list of internal plugins.
> If you need help doing this I can provide you some examples.

I would be very grateful.

towerz:
We have some changes planned in Clappr that aren't totally compatible with your changes. Please bear with me while I try to explain.

1 - About global plugin lists:
```javascript
Clappr.defaultPlugins.playback  = [...]
var player = new Clappr.Player({...})
```
This looks like what we'll use for plugin bundles, so I'm not totally against this. As I said before, what's part of the player core implementation should not be exposed, but if we need to expose something it will be part of the API or a plugin. So we would probably expose functions to allow this to work instead of setting properties directly.

2 - It doesn't seem comfortable, but as I told you, if you need to implement something more specific, you could have your own loader implementation and use it instead:
```javascript
var MyLoader = Clappr.Utils.extend(Clappr.Loader, {
  initialize: function(externalPlugins, playerId) {
    this.playbackPlugins = [Clappr.HTML5Video, Clappr.HTML5Audio, Clappr.FlasHLS, Clappr.HTMLImg, Clappr.NoOp]
    if (externalPlugins) {
      this.validateExternalPluginsType(externalPlugins)
      this.addExternalPlugins(externalPlugins)
    }
  }
});

var MyPlayer = Clappr.Utils.extend(Clappr.Player, {
  initialize: function() {
    this.loader = new MyLoader(this.options.plugins || {}, this.options.playerId);
  }
});

var player = new MyPlayer({
  baseUrl: '/latest',
  parentId: '#player-wrapper',
  mute: true,
  height: 360,
  width: 640
});

player.on(Clappr.Events.PLAYER_READY, function() {
  console.log(player.loader.playbackPlugins);
});
```
It could be even this implementation you wrote in this PR. The fact is that when you add a way to override internal plugins, there's no real distinction between internal or external plugins. Actually, one of our plans is to make all plugins external, and have them bundled in the default clappr package, something like clappr-all.min.js. The Clappr implementation would only include the HTML5 and Flash fallback playbacks, and no hardcoded plugins. This way you would be able to create your own bundle, so this PR would not be needed.

3 - We use plugin classes in the `plugins` parameter to allow loading external components. If you create a plugin named "my_plugin", you wouldn't be able to load it by name, only using its class:
```javascript
var MyPlugin = Clappr.Utils.extend(Clappr.UICorePlugin, { get name() { return "my_plugin"} });

  //...
  plugins: { core: ["my_plugin"] } // ???? How would that work?
});
```

About your other questions:

Unlike container/core plugins, playbacks can't be disabled, but you don't need to patch loader.js to control the playback lists. Instead, you could extend `Loader` and change its lists in the `initialize` method. It's not a hack. Something like the example on 2.

To disable plugins, you could just call `disable()` on them. Something like:
```javascript
player.getPlugin('stats').disable();
```

I noticed the NoOp plugin and Loader were not being exposed in the Clappr namespace, but it's already fixed, so all examples here would work.

avreg:
В Tue, 15 Dec 2015 13:26:53 -0800
Bruno Torres <notifications@github.com> пишет:

> This looks like what we'll use for plugin bundles, so I'm not totally against this. As I said before, what's part of the player core implementation should not be exposed, but if we need to expose something it will be part of the API or a plugin. So we would probably expose functions to allow this to work instead of setting properties directly.

Yes, but I am sure that the player core will still consist of a number of internal the most used plugins,
like as playback's (html5, flash), container (spinner, click-to-pause), maybe core(seektime).
If you remove this most used plugins from the player core,
Clappr will be similar to the constructor for advanced users.
If you do not remove this most used plugins from player core,
it is desirable to provide a simple and intuitive way to change a set of plugins.

 
> 2 - It doesn't seem comfortable, but as I told you, if you need to implement something more specific,
> you could have your own loader implementation and use it instead:
> ```javascript
> var MyLoader = Clappr.Utils.extend(Clappr.Loader, {
> var MyPlayer = Clappr.Utils.extend(Clappr.Player, {
> ```

> To disable plugins, you could just call `disable()` on them. Something like:
> ```javascript
> player.getPlugin('stats').disable();
> ```

Many thanks for these tips.
and other users can use them as a **temporary** solution (imho).

>   plugins: { core: ["my_plugin"] } // ???? How would that work?

Current plugins hangle disadvantages:
1) external plugin classes litter frontend app namespace;
2) internal plugin classes litter Clappr global variable;
3) uncomfortable and not work reliably with the (long) class names of plugins
instead short string names (and documented).

Please look at this idea below.

```
[CLAPPR]/base/plugin.js

  class Plugin extends BaseObject {
    static get pluginName() { throw "You must override this method in your plugin's class" }
    static get pluginType() { throw "You must override this method in your plugin's class" }
    get name() { return this.constructor.pluginName }
    get type() { return this.constructor.pluginType }
  }

[CLAPPR]/base/plugin_registar.js

  class PluginRegistar {
     constructor() {
       super()
       this._plugins = {
         playback: [],
         container: [],
         core: []
       }
    }

    function register(pluginClass) {
      // check type
      // append to this._plugins[type]
      // if global Clappr.plugingsOrder[type] != undefined
      //    - perhaps disable pluginClass
      //    - (re)sort this._plugins[type]
    }
    
    function getOrderedClassesList(type, override) {
      if (override) {
        // override and resort this._plugins[type] and return
      } else {
         return this._plugins[type]
      }
    }
  }

  var pluginRegistar = new PluginRegistar()
  let registerPlugin = function(pluginClass) {
    pluginRegistar.register(pluginClass)
  }

  export {registerPlugin, pluginRegistar}


[CLAPPR]/internal_plugins.js

  import {registerPlugin} from "base/plugin_registar.js"

  class SomeClapprInternalPlugin extends Plugin {
    static get pluginName() { return "html5" } // for example
    static get pluginType() { return Clappr.PLUGIN_TYPE_PLAYBACK }
  }

  registerPlugin(SomeClapprInternalPlugin)


[EXT.PLUGIN]/external_plugins.js

  class SomeClapprExternalPlugin extends Plugin {
    static get pluginName() { return "rtmp" } // or "rtmp-vendor" if there may be a few "rtmp" plugins
    static get pluginType() { return Clappr.PLUGIN_TYPE_PLAYBACK }
  }

  registerPlugin(SomeClapprExternalPlugin)


[CLAPPR]/components/loader/loader.js
  import {pluginRegistar} from "base/plugin_registar.js"
  
  class Loader {
    /***
    * customPluginsNames = options.plugins
    * new Clappr.Player({
    *    plugins: {
    *      playback: [],
    *      container: []
    *      core: []
    *    }
    * })
    */
    constructor(playerId, customPlugins) {
      ...
      this.playbackPlugins = pluginRegistar.getOrderedClassesList('playback', customPluginsNames['playback'] || {})
      this.containerPlugins = pluginRegistar.getOrderedClassesList('container', customPluginsNames['container'] || {})
      this.corePlugins = pluginRegistar.getOrderedClassesList('core', customPluginsNames['core'] || {})
      ...
  }


[CLAPPR]/main.js
  * Remove all internal plugin classes from export.
  * (optionally) Import pluginRegistar and add to export (Clappr.pluginRegistar).

```

towerz:
@avreg thanks for your insights. We'll take them into account when developing the plugin bundling feature.

avreg:
В Wed, 16 Dec 2015 05:27:06 -0800
Bruno Torres <notifications@github.com> пишет:

> @avreg thanks for your insights. We'll take them into account when developing the plugin bundling feature.

I hope so, and I wish you success.

PR #740, opened 2015-12-15
Show duration on seek time only when when live stream with dvr

tjenkinson:
This should mimic the behaviour of the current released version.

### DIFF
--- a/src/plugins/seek_time/seek_time.js
+++ b/src/plugins/seek_time/seek_time.js
@@ -25,7 +25,7 @@ export default class SeekTime extends UICorePlugin {
   }
   get mediaControl() { return this.core.mediaControl }
   get mediaControlContainer() { return this.mediaControl.container }
-  get durationShown() { return this.mediaControlContainer.getPlaybackType() !== Playback.LIVE || this.mediaControlContainer.isDvrInUse() }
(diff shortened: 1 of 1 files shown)


src/plugins/seek_time/seek_time.js
```diff
@@ -25,7 +25,7 @@ export default class SeekTime extends UICorePlugin {
   }
   get mediaControl() { return this.core.mediaControl }
   get mediaControlContainer() { return this.mediaControl.container }
-  get durationShown() { return this.mediaControlContainer.getPlaybackType() !== Playback.LIVE || this.mediaControlContainer.isDvrInUse() }
+  get durationShown() { return this.mediaControlContainer.getPlaybackType() === Playback.LIVE && this.mediaControlContainer.isDvrEnabled() }
```
leandromoreira: but how about when it's `Playback.VOD`?

tjenkinson: When it's vod the duration and current time is shown in the media control
image

leandromoreira: my bad!!!

towerz:
Nice catch, @tjenkinson!

PR #741, opened 2015-12-15
Actual server time now only used when stream is live and has dvr

### LINKED ISSUE #714: If 'actualLiveTime' feature enabled, but stream type changes to "VOD", default back to normal timing
The 'actualLiveTime' feature works great when a stream is live, but if a stream ends and goes from being live to VOD (which can happen with a hls EVENT playlist) then it stops working as it relies on the latest part of the content being at the live point.

It would be great if this feature could automatically disable itself if the media is not a live stream, or the live stream stops being live, and revert back to displaying the seek time from the beginning of the content.

tjenkinson:
So if a live stream stops being live the time will change back to counting from 0.

Closes #714


towerz:
Awesome, thanks @tjenkinson!

PR #744, opened 2015-12-16
fix browser.isChrome

avreg:


towerz:
Nice catch, thanks @avreg!

PR #747, opened 2015-12-20
HLSJS now initialised immediately

tjenkinson:
So that PLAYER_READY can fired before the users requests to play the content.

### DIFF
--- a/src/playbacks/hls/hls.js
+++ b/src/playbacks/hls/hls.js
@@ -41,7 +41,6 @@ export default class HLS extends HTML5VideoPlayback {
   setupHls() {
     this.hls = new HLSJS(this.options.hlsjsConfig || {})
     this.hls.on(HLSJS.Events.MEDIA_ATTACHED, () => this.hls.loadSource(this.options.src))
-    this.hls.on(HLSJS.Events.MANIFEST_PARSED, () => { this.options.autoPlay && this.play() })
     this.hls.on(HLSJS.Events.LEVEL_LOADED, (evt, data) => this.updatePlaybackType(evt, data))
     this.hls.on(HLSJS.Events.LEVEL_UPDATED, (evt, data) => this.updateDuration(evt, data))
     this.hls.on(HLSJS.Events.LEVEL_SWITCH, (evt,data) => this.onLevelSwitch(evt, data))
@@ -122,6 +121,12 @@ export default class HLS extends HTML5VideoPlayback {
       delete this.hls
       this.trigger(Events.PLAYBACK_STOP)
     }
+  }
+
+  render() {
+    var a = super.render()
+    this.ready()
+    return a
   }
 
   updatePlaybackType(evt, data) {
--- a/src/playbacks/hls/hls.js
+++ b/src/playbacks/hls/hls.js
@@ -124,9 +124,9 @@ export default class HLS extends HTML5VideoPlayback {
   }
 
   render() {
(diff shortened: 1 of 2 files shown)


towerz:
@tjenkinson I don't think this is the best solution, since hls.js starts preloading the video, as if playback has been started. We avoided initializing the hls object immediately to avoid unnecessary requests. Unless preloading the video is an options and the user chooses this, I wouldn't do it this way.

About the poster disappearing in the first bufferfull event, it's because the bufferfull event should only be triggered when loading the video, not when preloading. I would let this change on the poster plugin to a later time, since we want to avoid the poster to manage it's own state, but instead being controlled by the container.

Perhaps the right approach is to consider the playback ready after rendering, or if preload is enabled, after preloading. The buffering event shouldn't be triggered on preload, unless autoPlay is enabled.

towerz:
Sounds ok for me. Perhaps we could tranlslate the `preload` option that we use on the html5_video playback to `autoStartLoad`, or add to as an exposed HLS option. By default, there should be no preload, so this proposed solution should work ok.

towerz:
@tjenkinson why don't we trigger the ready state only after `render` instead of in the constructor? As far as I know, the listeners are always added before rendering, so there's no risk of the event being triggered without listeners. Actually this would be more accurate, because a video element should only be ready after it's added to the DOM.

src/playbacks/hls/hls.js
```diff
@@ -124,6 +123,12 @@ export default class HLS extends HTML5VideoPlayback {
     }
   }
 
+  render() {
+    var a = super.render()
```
leandromoreira: what's `a` ? is it super.element (video tag element) ? If yes, I think we should name this to `videoElement` or something that show intention

tjenkinson: It's 'this' from Html5Video, which I think becomes the same 'this' from Hls?

leandromoreira: :) I think you're right! maybe we can get rid of that `a`, my observation was about naming that `a` could mislead or make us to go to `HTML5Video` to understand what's `a`. I think, `return this` is best.

```javascript
render() {
  super.render()
  this.ready()
  return this
}
```

tjenkinson: Yep will tweak it

src/playbacks/html5_video/html5_video.js
```diff
@@ -269,8 +269,11 @@ export default class HTML5Video extends Playback {
   }
 
   ready() {
-    this.trigger(Events.PLAYBACK_READY, this.name)
+    if (this.isReadyState) {
+      return
```
flavioribeiro: why we need this? ready() is being called more than once?

tjenkinson: Because it looks like it could be called [here](https://github.com/clappr/clappr/blob/fix-hlsjs-ready-event/src/playbacks/html5_video/html5_video.js#L302) as well as [here](https://github.com/clappr/clappr/blob/fix-hlsjs-ready-event/src/playbacks/hls/hls.js#L128)

tjenkinson: And [here](https://github.com/clappr/clappr/blob/fix-hlsjs-ready-event/src/playbacks/html5_video/html5_video.js#L48)

leandromoreira:
Thanks @tjenkinson

PR #748, opened 2015-12-21
Documenting playback events

tjenkinson:
Started adding missing event docs.

There are a few TODOs for playback events I'm not sure about.

Also the `FRAGMENT_LOADED` event is specific to hlsjs at the moment, and it's data comes straight from the hlsjs event. Should this be standardised?


leandromoreira:
yep I don't see `FRAGMENT_LOADED` for any other playback type... and how is it useful to the plugins system?

towerz:
Thanks @tjenkinson, great job.

PR #749, opened 2015-12-21
Refactor poster plugin

tjenkinson:
No longer listens to `CONTAINER_STATE_BUFFERING` or `CONTAINER_STATE_BUFFERFILL` and removes some duplicated code, as I think the same is being achieved now without those events?

Tested with a hls stream and mp4 and seems to be working.

Also this fixes a bug where the poster was hidden incorrectly when #747 is being used (although that pr is currently out of date).

### DIFF
--- a/src/plugins/poster/poster.js
+++ b/src/plugins/poster/poster.js
@@ -15,6 +15,7 @@ import $ from 'clappr-zepto'
 export default class PosterPlugin extends UIContainerPlugin {
   get name() { return 'poster' }
   get template() { return template(posterHTML) }
+  get shouldRender() { return this.container.playback.name !== 'html_img'}
 
   get attributes() {
     return {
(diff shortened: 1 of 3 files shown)


towerz:
Hey @tjenkinson, this refactor seems a good idea, but there are two main points that I would like to suggest that you change.

First, there's the `data-clickable` attribute. Instead of doing this, I'd rather add a `clickable` class. I'm not a big fan of using attributes, specially with data prefix, to control applying styles. Using data attributes is usually done for setting actual data to be used by whatever component is accessing the element. In Clappr, the data attributes are actually legacy, we used in early prototypes to apply individual styles to elements depending on which player instance they're rendered into (the player id was a value applied to these attributes). The data attributes have been since replaced by classes, and are being used only to increase css priority, but I think they will be removed soon.

The other one is the use of `setTimeout(fn, 0)`. We are using `process.nextTick` instead of this, to keep Clappr compatible with the node.js default api. There's already a browser shim to be used in Clappr, so we'd rather use this. When we changed the other `setTimeout` calls, we must have missed this one.

As I told you before, I didn't want the poster to have that much logic in it. It shouldn't be responsible for keeping state in it, but I don't think it's the right time to change this, so we can go on with this refactor. Could you please change these two issues I pointed?

leandromoreira:
It looks great to me :) thoughts @clappr/core ?

leandromoreira:
I'm sorry man =) and once again thank you very much.

PR #754, opened 2015-12-28
adds hasEnded() method to playbacks, container, and player

tjenkinson:
Tested in Chrome with `Html5Video`, `FlasHLS` and `HLS` playbacks.

### DIFF
--- a/src/base/playback.js
+++ b/src/base/playback.js
@@ -67,6 +67,15 @@ export default class Playback extends UIObject {
   }
 
   /**
+   * checks if the playback has ended.
+   * @method hasEnded
+   * @return {Boolean} `true` if the current playback has ended, otherwise `false`
+   */
+  hasEnded() {
+    return false
+  }
+
+  /**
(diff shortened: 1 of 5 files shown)


leandromoreira:
I'm in favor of another property (like we did for ready) or we should change everything to methods :P what do you think @clappr/core ?

```javascript
get ended() {  return this._ended }
//or
isEnded() { return this._ended }
```

I vote in favor of property because the states (ready, ended, paused ...) sound more like a property than a method/function.

flavioribeiro:
just curious about the use case of this method, why not listen to `PLAYBACK_ENDED`?

leandromoreira:
@clappr/core are we going to use `attribute` or `method` to cover this ? **hasEnded** (or ended)

leandromoreira:
Cool @tjenkinson are you going to rebase or do another PR for `ended` instead of `hasEnded()`

leandromoreira:
I'll close, thanks once more :)

PR #759, opened 2016-01-08
flashls: update FlasHLS to 0.4.4.17

me-vlad:
flashls: update HLSPlayer.swf to 0.4.10 (flashls 0.4.4.17) and flashlsChromeless.swf to 0.4.4.17
flashls: Add maxSkippedFragments setting to define max skipped fragments before throwing an error


leandromoreira:
Thanks :)

PR #765, opened 2016-01-13
Change isReady() to isReady in Core and Player

tjenkinson:
To be consistent with other places. This changes the api, but the `isReady()` method was also only added recently.


leandromoreira:
Thanks

PR #766, opened 2016-01-14
poster-icon: paint poster icon with custom colors (fixes #763)

### LINKED ISSUE #763: Mediacontrol color for buttons does not change poster-icon color
Large play button (poster-icon) is white even though button has a color set to it. Just seems unintuitive that the first and largest button doesn't match the rest

button color

This seems intentional but should be an easy change to allow more options for the users.

It is explicitly set to "color: white;" and instead could use the entered value.

Very similar to clappr#224

hounvs:


flavioribeiro:
Thanks @hounvs! I hope you can contribute even more to the project! :beers:

PR #767, opened 2016-01-14
Fix so that cursor is hidden when idle in fullscreen

### LINKED ISSUE #753: Cursor remains visible as "pointer" when full screen (chrome)
The "nocursor" class that gets added on the player is getting overridden by the "pointer-enabled" class on the container.

tjenkinson:
(fixes #753)


leandromoreira:
THANK YOU SO MUCH :) 
I always want that :) :beers:

PR #772, opened 2016-01-16
Ensure poster play button remains visible if autoplay doesn't work

tjenkinson:
on iOS devices for example


leandromoreira:
Thanks :+1:

PR #773, opened 2016-01-17
Fix issues when player loaded with no source

bikegriffith:
1. `requestAnimationFrame` leaking and killing CPU on canvas noise generator.  I was trying to re-use a single player instance across a playlist of videos (some of which were faulty during testing), and I found my CPU being pegged constantly.  Profiling found `noise()` to be the culprit.  Using `cancelAnimationFrame` (already polyfilled in utils) to take care of this.
2. `player.destroy()` fails unbinding kibo events - looks like the kibo safety check was missed in a few spots, so moved it closer to where it was needed

### DIFF
--- a/src/components/media_control/media_control.js
+++ b/src/components/media_control/media_control.js
@@ -520,9 +520,7 @@ export default class MediaControl extends UIObject {
   }
 
   bindKeyEvents() {
-    if (this.kibo) {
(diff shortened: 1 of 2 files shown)


leandromoreira:
nice catch @bikegriffith thank u very much, we hope to see more pr's from u :beers:

PR #778, opened 2016-01-19
Make it easier to create custom MediaControl

bikegriffith:
* Add overridable `stylesheet` property to base MediaControl rather than creating in constructor
* Update README with sample of how to do it
* Add unit test


leandromoreira:
Looks great to me too! thank you @bikegriffith

Issue #197, opened 2014-10-09
Create default siding for audio only streams


flavioribeiro:
When clappr is playing HLS audio-only streams or even mp3 musics it shows nothing on the screen. We could put a siding to notice the user that the stream being played is audio only.

PR #783, opened 2016-01-20
Allow poster to stay open for audio-only broadcasts

bikegriffith:
The use-case is an audio-only HLS stream, which ends up using the `<video>` playback.  We want to have the poster stick around in this case, even though the source is playing.

### DIFF
--- a/src/plugins/poster/poster.js
+++ b/src/plugins/poster/poster.js
@@ -107,6 +107,11 @@ export default class PosterPlugin extends UIContainerPlugin {
     }
   }
 
+  shouldHideOnPlay() {
+    // Audio broadcasts should keep the poster up; video should hide poster while playing.
+    return !((this.container.playback.name == 'html5_audio') || this.options.audioOnly);
+  }
+
   update() {
     if (!this.shouldRender) {
       return
@@ -118,7 +123,9 @@ export default class PosterPlugin extends UIContainerPlugin {
     }
     else {
       this.container.enableMediaControl()
-      this.$el.hide()
+      if (this.shouldHideOnPlay()) {
+        this.$el.hide()
+      }
     }
   }
 
--- a/README.md
+++ b/README.md
@@ -186,7 +186,10 @@ Put `watermark: http://url/img.png` on your embed parameters to automatically ad
 ```
 
 ##### Poster
-Define a poster by adding `poster: http://url/img.png` on your embed parameters. It will appear after video embed, disappear on play and go back when user stops the video.
+Define a poster image by adding `poster: 'http://url/img.png'` on your player options. It will appear after video embed, disappear on play and go back when user stops the video. For audio broadcasts, the poster stays up while playing.
+
+##### Audio Only Hint
+Some audio-only sources (e.g. HLS) cannot be easily detected as such; for that you can add `audioOnly: true` to the options so clappr knows to treat the source as such.
 
 ##### Stats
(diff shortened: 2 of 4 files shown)


src/plugins/poster/poster.js
```diff
@@ -108,6 +108,12 @@ export default class PosterPlugin extends UIContainerPlugin {
   }
 
   update() {
+    let hideOnPlay = true;
+    if (this.options.posterConfig && typeof(this.options.posterConfig.hideOnPlay) != 'undefined') {
+      // There are circumstances to keep poster around when we'd normally hide it
+      // (e.g. audio-only source inside video container format)
+      hideOnPlay = this.options.posterConfig.hideOnPlay;
+    }
```
tjenkinson: This can be simplified to:

let hideOnPlay = !!(this.options.posterConfig && this.options.posterConfig.hideOnPlay);

bikegriffith:
@tjenkinson Do you know how to manually specify the playback?  Adding `playback: Clappr.HTML5Audio` doesn't appear to be working.

src/plugins/poster/poster.js
```diff
@@ -15,7 +15,8 @@ import $ from 'clappr-zepto'
 export default class PosterPlugin extends UIContainerPlugin {
   get name() { return 'poster' }
   get template() { return template(posterHTML) }
-  get shouldRender() { return this.container.playback.name !== 'html_img'}
+  get shouldRender() { return this.container.playback.name !== 'html_img' }
+  get shouldHideOnPlay() { return this.container.playback.name !== 'html_audio' }
```
bikegriffith: This isn't reliably working for me because I can't force HTML5Audio playback for a HLS audio-only stream. My last commit (manually specifying poster auto-hide behavior in options) worked fine for my use case, although this would be more elegant.

tjenkinson: Ah yes I'd forgotten there's only the video hls playback.

Hls audio only playlists do exist though so maybe there should also be a HlsAudio playback? It gets complicated though because I think the extension would always be 'm3u8' so it would need more than an extension match :/

leandromoreira:
Thank you very much @bikegriffith :)

I think since it's almost (in an easy and clean way) impossible for Clappr to figure out whether it is playing and audio or video.

You could inform that on player's setup. Maybe by adding a plugin to handle that, @flavioribeiro already created an issue for something similar that https://github.com/clappr/clappr/issues/197 there are great ideias there.

What do you think @bikegriffith and @tjenkinson ?

```javascript
new  Clappr.Player({plugins: [ClapprAudioUI]}
```

bikegriffith:
@leandromoreira Yeah, this is tricky.  I wonder if just using an option like I had in https://github.com/bikegriffith/clappr/commit/0258ed50a02731287127c7d216daff74d3f4ff71 is good enough ... it puts the separation of concerns in roughly the same spot (user needs to inform player instance what to expect).

leandromoreira:
Maybe a hint `this.options.audioOnly` instead of a poster config `this.options.posterConfig.hideOnPlay` can be useful in other places too.

bikegriffith:
@leandromoreira I like that idea! This is feeling pretty good to me...

leandromoreira:
@bikegriffith thanks ;) can you squash the commits into a few? also for the documentation part I think you might add `audioOnly` as whole new option, after that we can merge it.

bikegriffith:
@leandromoreira rebased/squashed everything and updated README.md.  let me know if this looks OK or if you want anything else changed.  Thanks!

leandromoreira:
Thanks a lot @bikegriffith :)

PR #787, opened 2016-01-21
Refactored BUFFERING and BUFFERFULL events and added 'buffering' property

### LINKED ISSUE #785: FF 43.0.4 buffering icon problem (spinning too much after play or all time)
After click play spinning all the time.
After manual seek, icon buffering (spinning) disappear, what is ok behavior .

HLS VOD multibitrate
hlsjs
LevelSelector plugin
WINDOWS 10
on Chrome 47 and IE Edge behavior OK

tjenkinson:
`PLAYBACK_BUFFERING` and `PLAYBACK_BUFFERFULL` are now only fired whenever the new 'buffering' state changes.

The `PLAYBACK_BUFFERING` event is fired when loading has started the tech isn't happy that it's got enough content to play smoothly. The `PLAYBACK_BUFFERFULL`event is fired when the tech is happy it's got enough content to play smoothly.

The `readyState` in chrome (didn't test firefox) remained at `HAVE_ENOUGH_DATA` even when it had was stuttering/stalled, so there's also a separate check to determine if the playhead is actually moving.
'PLAYBACK_BUFFERING' is triggered if the playhead isn't moving but the tech is stating that it should be playing.

`PLAYBACK_READY` is now fired when rendering finishes, as at this point`'play()` can actually be called successfully and it means it's the same for `html5_video` as `hls`.

Tested hlsjs mp4 in internet explorer, edge, firefox and chrome and seems to be fine. Also tested flashls in chrome as well.

(Fixes #785)

### DIFF
--- a/src/playbacks/html5_video/html5_video.js
+++ b/src/playbacks/html5_video/html5_video.js
@@ -80,8 +80,7 @@ export default class HTML5Video extends Playback {
     super(options)
     this.loadStarted = false
     this.playheadMoving = false
-    this.playheadMovingTimeOnCheck = null
-    this.playheadMovingTimer = setInterval(this.determineIfPlayheadMoving.bind(this), 500)
+    this.playheadMovingTimer = null
     this.options = options
     this.setupSrc(options.src)
     this.el.loop = options.loop
@@ -154,6 +153,7 @@ export default class HTML5Video extends Playback {
   }
 
   play() {
+    this.startPlayheadMovingChecks()
     if (!this.loadStarted && this.el.preload === 'none') {
       this.loadStarted = true
       this.handleBufferingEvents()
@@ -168,6 +168,7 @@ export default class HTML5Video extends Playback {
   stop() {
     this.pause()
     this.el.currentTime = 0
+    this.stopPlayheadMovingChecks()
     this.handleBufferingEvents()
     this.trigger(Events.PLAYBACK_STOP)
   }
@@ -194,6 +195,24 @@ export default class HTML5Video extends Playback {
 
   get isReady() {
     return this.isReadyState
+  }
+
+  startPlayheadMovingChecks() {
+    if (this.playheadMovingTimer !== null) {
+      return
+    }
+    this.playheadMovingTimeOnCheck = null
+    this.determineIfPlayheadMoving()
+    this.playheadMovingTimer = setInterval(this.determineIfPlayheadMoving.bind(this), 500)
+  }
+
+  stopPlayheadMovingChecks() {
+    if (this.playheadMovingTimer === null) {
+      return
+    }
+    clearInterval(this.playheadMovingTimer)
+    this.playheadMovingTimer = null
+    this.playheadMoving = false
   }
 
   determineIfPlayheadMoving() {
@@ -254,7 +273,6 @@ export default class HTML5Video extends Playback {
     this.stop()
     this.el.src = ''
     this.src = null
-    clearInterval(this.playheadMovingTimer)
     this.$el.remove()
   }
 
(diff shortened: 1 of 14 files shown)


leandromoreira:
good Lord, aren't you a trooper ?! :) great work, I'll find a time to review it but anyway thanks for this load of work.

src/playbacks/html5_video/html5_video.js
```diff
@@ -62,19 +65,37 @@ export default class HTML5Video extends Playback {
...
+  get buffering() {
+    return this.bufferingState
+  }
+
   constructor(options) {
     super(options)
+    this.loadStarted = false
+    this.playheadMoving = false
+    this.playheadMovingTimeOnCheck = null
+    this.playheadMovingTimer = setInterval(this.determineIfPlayheadMoving.bind(this), 500)
```
leandromoreira: I think we can get rid off bind:

```javascript
setInterval(() => this.determineIfPlayheadMoving(), 500)
```

src/playbacks/html5_video/html5_video.js
```diff
@@ -261,19 +309,8 @@ export default class HTML5Video extends Playback {
...
-      this.trigger(Events.PLAYBACK_BUFFERING, this.name)
-      this.buffering = true
-    } else if (this.buffering) {
-      this.trigger(Events.PLAYBACK_BUFFERFULL, this.name)
-      this.buffering = false
-    }
-  }
-
   typeFor(src) {
-    return (src.indexOf('.m3u8') > 0) ? 'application/vnd.apple.mpegurl' : 'video/mp4'
```
leandromoreira: Why did you change this?

src/playbacks/html5_video/html5_video.js
```diff
@@ -166,51 +196,65 @@ export default class HTML5Video extends Playback {
...
-  error(event) {
+  onError(event) {
     this.trigger(Events.PLAYBACK_ERROR, this.el.error, this.name)
   }
 
   destroy() {
     this.stop()
     this.el.src = ''
+    this.src = null
+    clearInterval(this.playheadMovingTimer)
```
leandromoreira: Maybe we should `clearInterval` at stop too otherwise it'll keep checking bufferstate... this would require to start it again on play (if it's not already started).

I'm thinking that after a media is done, it'll keep triggering `BUFFERFULL`

src/playbacks/html5_video/html5_video.js
tjenkinson: because `typeFor` was only called in `html5_video`, not `hls`, meaning it would always be `video/mp4`. With `hls` no type gets added as this was for the `<source>` tags which aren't used for the hls playback.

src/playbacks/html5_video/html5_video.js
tjenkinson: Cool will do this :)

src/playbacks/html5_video/html5_video.js
leandromoreira: Android/Ios devices uses hls natively on vide tag.

src/playbacks/html5_video/html5_video.js
tjenkinson: Sounds good. Yes I'll move it to start on play and stop on stop.

It shouldn't do because in `handleBufferingEvents()` it ensures that the events must alternate.

src/playbacks/html5_video/html5_video.js
tjenkinson: Oh yeh. I was thinking the `hls` playback was still the one that would be used. Will put that back

src/playbacks/html5_video/html5_video.js
leandromoreira: If you move that for `onPlay` and `onStop` you still need to ensure that buffering state (handleBuffering) is changing even before it plays any media.

leandromoreira:
Great load of work :) thank you much @tjenkinson an awesome work! :clap: :clap: :clap: :clap:

src/playbacks/html5_video/html5_video.js
tjenkinson: It was actually needed here because the scope of the `setInterval` callback isn't the object.

leandromoreira: :+1:

PR #810, opened 2016-01-28
Poster play icon now hidden during buffering

tjenkinson:
Should also fix bug where play icon not visible on devices where autoplay is blocked, when autoplay has been requested. Not been able to test this on a mobile yet though.

### DIFF
--- a/src/plugins/poster/poster.js
+++ b/src/plugins/poster/poster.js
@@ -42,12 +42,8 @@ export default class PosterPlugin extends UIContainerPlugin {
     this.listenTo(this.container, Events.CONTAINER_STOP, this.onStop)
     this.listenTo(this.container, Events.CONTAINER_PLAY, this.onPlay)
     this.listenTo(this.container, Events.CONTAINER_ENDED, this.onStop)
-    if (this.container.buffering) {
-      process.nextTick(() => this.onBuffering())
-    }
-    else {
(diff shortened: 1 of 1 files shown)


src/plugins/poster/poster.js
```diff
@@ -42,12 +42,8 @@ export default class PosterPlugin extends UIContainerPlugin {
     this.listenTo(this.container, Events.CONTAINER_STOP, this.onStop)
     this.listenTo(this.container, Events.CONTAINER_PLAY, this.onPlay)
     this.listenTo(this.container, Events.CONTAINER_ENDED, this.onStop)
-    if (this.container.buffering) {
-      process.nextTick(() => this.onBuffering())
```
leandromoreira: Did you test with FlasHLS? ( I think this `process.nextTick` is to handle flash)

tjenkinson: I didn't, but this was something I put in a couple of commits ago, and the same should be being achieved now with `update()` being called at the end of render, and the `.buffering` check in render anyway

leandromoreira: Cool :+1:  thanks once again mister

tjenkinson: Technically `buffering` doesn't need to be checked. But it means the play button disappears if the loading indicator is on, since they both check `buffering`, instead of play still being visible over the top if the user hasn't clicked it yet.

PR #812, opened 2016-01-29
Fix three spinner bounce buffer handling

tjenkinson:
Can't assume the playback is still buffering a tick after the point the plugin loads.

Was causing bug when native hls used. Loading indicator was sticking on screen until play clicked on iOS. Think it might be that on ios the buffering doesn't start until play is requested.


leandromoreira:
@tjenkinson THANK YOU VERY MUCH! :clap: :clap: :clap: :clap: Well Done :clap: :clap: :clap:

PR #814, opened 2016-01-30
Html5Video playback now only buffering after first frame loaded, or user action

tjenkinson:
This fixes issues with devices which disable autoplay, like ios. The "canplay" event fires for these devices before user interaction is required. "loadeddata", fires after user interaction has been provided.

Therefore "loadeddata" (first frame loaded) is now used to signify that loading has started. However there's also a "waiting" event, which if triggered will also signify that loading has started. The waiting event seems to be fired after user interaction when waiting is required before playback can occur. I have checked as well and on ios programmatically calling a method like play() or seek() does not trigger the 'waiting' event. I guess iOS is just ignoring these requests completely.

This means if autoplay is not requested in the player options, once the user has clicked play the buffering state will immediately be true, because "waiting" is fired immediately. If autoplay is requested in the player options then buffering will only start when the first frame is loaded (as 'waiting' won't be fired), but this shouldn't be a significant delay. If the user was to request play during the period before the first frame was loaded, then 'waiting' would be fired, and the buffering state would be entered immediately.

### DIFF
--- a/src/playbacks/html5_video/html5_video.js
+++ b/src/playbacks/html5_video/html5_video.js
@@ -38,13 +38,13 @@ export default class HTML5Video extends Playback {
 
   get events() {
     return {
-      'canplay': 'handleBufferingEvents',
+      'canplay': 'onCanPlay',
       'canplaythrough': 'handleBufferingEvents',
       'durationchange': 'onDurationChange',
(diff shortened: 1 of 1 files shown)


leandromoreira:
pretty neat congrats again :)

PR #815, opened 2016-01-31
Set buffering to false when Html5Video is stopped

tjenkinson:
This fixes a bug which would cause the play button not to reappear when clicking stop on a live stream (because hlsjs is detached, but the video element still thinks it's buffering).

I think this is more of a simple fix for now. When hlsjs is detached I don't think `buffering` being true really makes sense, but not sure at the moment how to make it `false`. Maybe on stop() the video/audio element should be destroyed and recreated so it starts fresh?


leandromoreira:
I'd assume `stop` being a safe place, lgtm

Issue #788, opened 2016-01-22
FlasHLS playback plugin not working on IE10


michaelcunningham19:
The implementation of FlasHLS is failing on IE10 on Windows 7, and Windows 8

Environment:
Windows 7
Windows 8
IE10
http://cdn.clappr.io

Using these values:
```
var playerElement = document.getElementById("player-wrapper");

var player = new Clappr.Player({
  source: 'http://www.streambox.fr/playlists/x36xhzz/x36xhzz.m3u8',
  baseUrl: '/latest',
  poster: 'http://clappr.io/poster.png',
  mute: true,
  height: 360,
  width: 640,
  plugins: {
    playback: [Clappr.FlasHLS]
  }
});

player.attachTo(playerElement);
```

The following error occurs:
```
external undefined plugin on playback array
```

It's also worth noting that on this page:
http://www.flashls.org/latest/examples/chromeless/
With the URL:
http://www.streambox.fr/playlists/x36xhzz/x36xhzz.m3u8

Works perfectly fine in IE10.

leandromoreira:
@michaelcunningham19 try this:

```javascript
var player = new Clappr.Player({
  source: 'http://www.streambox.fr/playlists/x36xhzz/x36xhzz.m3u8',
  baseUrl: '/latest',
  poster: 'http://clappr.io/poster.png',
  mute: true,
  height: 360,
  width: 640,
  plugins: [Clappr.FlasHLS]
});
```

michaelcunningham19:
@leandromoreira 

Tested your suggestion in IE10,

No errors were thrown however upon clicking the play icon, nothing happened.

leandromoreira:
@michaelcunningham19 try to enable log level debug 

```javascript
// before instantiate the player

Clappr.Log.setLevel(0);
//..
```

And try paste the log in here.

michaelcunningham19:
@leandromoreira 

On the environment:
IE10 on Windows 8 (and also 7) on http://cdn.clappr.io

With the following config:
```
var playerElement = document.getElementById("player-wrapper");

Clappr.Log.setLevel(0);

var player = new Clappr.Player({
  source: 'http://www.streambox.fr/playlists/x36xhzz/x36xhzz.m3u8',
  baseUrl: '/latest',
  poster: 'http://clappr.io/poster.png',
  mute: true,
  height: 360,
  width: 640,
  plugins: [Clappr.FlasHLS]
});

player.attachTo(playerElement);
```

[...]

PR #816, opened 2016-02-01
Added support for ES6 methods for older browsers

michaelcunningham19:
This PR fixes the error:

```
 %c[error]color: #ff0000;font-weight: bold; font-size: 13px;error on eventcore:containers:createdtrigger-TypeError: Object doesn't support property or method 'find' 
```

Which occurs on any browser which does not support the native ES6 method 'find' on the Array prototype object. (IE10 and IE11 for example)

Adding the official _babel-polyfill_ package solves this from occurring.

See more info: https://github.com/clappr/clappr/issues/788


michaelcunningham19:
@tjenkinson 

It polyfills the methods in the user's browsers.

I agree it would be nice to have it handled in the build process, but following the official usage instructions it says to simply 'require/import' at the top of the main entry point.

leandromoreira:
Thanks :) @michaelcunningham19

Can you remove `dist/` from your commits?

We @clappr/core also need to check the final weight added by this polyfill, maybe it's easier to use `https://lodash.com/docs#find` (that we already have) or even to provide a pollyfill using this or yet use the [mozilla polyfill](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find)

leandromoreira:
I prefer to fix this instead of adding another dependency at `src/plugins/sources.js`

```diff
+import find from 'lodash.find'

-var firstValidSource = this.core.containers.find((container) => container.playback.name !== 'no_op') || this.core.containers[0]
+var firstValidSource = find(this.core.containers((container) => container.playback.name !== 'no_op') || this.core.containers[0]
```

What do you think?

michaelcunningham19:
@leandromoreira 
That's a good catch, I didn't think of using lodash.find here!

I switched it up to use that instead rather than the polyfill.

@tjenkinson 
IE10 doesn't support any ES6 features as far as I know, so there could be more things like this pop up in the future, but for now I agree with not bringing in another dependency.

leandromoreira:
@michaelcunningham19 do another PR fixing the find with load.find :)

PR #818, opened 2016-02-01
Switched to lodash 'find' from native ES6 'find' for older browser support

michaelcunningham19:
This PR fixes the error:

```
 %c[error]color: #ff0000;font-weight: bold; font-size: 13px;error on eventcore:containers:createdtrigger-TypeError: Object doesn't support property or method 'find'
```

Which occurs on any browser which does not support the native ES6 method 'find' on the Array prototype object. (IE10 and IE11 for example)

Per discussion with @leandromoreira this is a new PR to fix the issue described in https://github.com/clappr/clappr/pull/816

### DIFF
--- a/src/plugins/sources.js
+++ b/src/plugins/sources.js
@@ -1,6 +1,7 @@
 import CorePlugin from 'base/core_plugin'
 
 import Events from 'base/events'
+import find from 'lodash.find'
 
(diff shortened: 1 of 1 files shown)


flavioribeiro:
Thanks, @michaelcunningham19!

PR #820, opened 2016-02-01
Removed FlashLS object element type attribute for legacy IE

michaelcunningham19:
This PR fixes the issue where IE10 would not play a video through FlashLS with no errors being outputted upon clicking the play icon. See related issue: https://github.com/clappr/clappr/issues/788

The final issue turned out to be that IE10 does not load a Flash file properly if there is a classid attribute as well as a 'type' attribute.

See this link for more information:
http://blogs.adobe.com/flashplayer/2015/10/flash-player-guidance-for-internet-explorer-11-and-microsoft-edge.html


leandromoreira:
Thank you so much @michaelcunningham19 ;) :clap: :clap: :clap: :clap: :clap: :clap: :clap:

Issue #793, opened 2016-01-25
shrinkwrap causing my linux CI build to fail


jfairley:
I'm trying to get upgraded to the latest version of clappr, and I've run into another build issue...

I've done a little research already to track down what is going on. Apparently when shrinkwrap runs, it takes all optional dependencies and dev dependencies (even dependencies of your dependencies) and dumps them all into "dependencies".

My specific issue seems to be with an "optional" dependency that is no longer marked as optional. When my CI machine tries to install [fsevents](https://www.npmjs.com/package/fsevents) on linux, everything bombs, because shrinkwrap implies that this dependency is required.

```
npm ERR! notsup Unsupported
npm ERR! notsup Not compatible with your operating system or architecture: fsevents@1.0.6
npm ERR! notsup Valid OS:    darwin
npm ERR! notsup Valid Arch:  any
npm ERR! notsup Actual OS:   linux
npm ERR! notsup Actual Arch: x64
```

Others have complained about this issue. It's [on the npm roadmap](https://github.com/npm/npm/wiki/Roadmap-area-of-focus%3A-bundling), but a fix is not yet implemented.
- https://github.com/npm/npm/issues/8707
- https://github.com/npm/npm/issues/2679

The only "solution" I've seen is that people stop using shrinkwrap altogether. :frowning: 

Has anyone else seen this issue? Do you build or install clappr on linux?

tjenkinson:
Yep I have the same issue on windows now, and removing the shrinkwrap file fixes it. Thanks for taking the time to research the issue. Interestingly `npm install` still runs fine on travis though.

I guess one solution would be to remove the shrinkwrap file from your copy during your build process?
The other looks like it would be scrap it all together until a nice solution is implemented. Optional dependencies becoming required in the shrinkwrap does sound like a flaw.

Are you able to ignore the shrinkwrap file during your build process?

jfairley:
Here is my temporary workaround:

1. `.gitignore`
  ```
  !node_modules/.hooks/*
  ```

2. `node_modules/.hooks/preinstall`
  ```bash
  #!/bin/sh
  if [[ `pwd` = */clappr ]]; then
      rm npm-shrinkwrap.json
  fi
  ```

:grin:

jfairley:
I had another idea based on how I release one of my projects.

[...]

PR #822, opened 2016-02-03
update release process to deploy only the dist directory

jfairley:
This changes the npm release process to deploy only the `dist` directory rather than the entire repo, as suggested [in #793](https://github.com/clappr/clappr/issues/793#issuecomment-179400290).

__This would do two things:__

1. Removes the unnecessary "npm install" that runs on a client machine when clappr is declared as a dependency. This speeds up user install time, as well as gets around the issue I brought up in #793.
1. Reduces the overall size of the package distributed to NPM (speeding up user download time).

If someone wants to run tests or make a custom build, they can always clone the repo. IMO, it's strange to do this inside `node_modules/clappr` anyway.

__What did I change?__

1. Copy `package.json` into `dist` and slim it down by removing "dependencies" and "devDependencies".
1. Update the `bump` script to version `dist/package.json` and publish from the `dist` directory.
1. Update `webpack.config.js` to not remove `dist/package.json` when it does the "clean".
1. Removed `.npmigore`, because it is no longer needed.

<br>

Lemme know what you think.


jfairley:
Coverage dropped with config changes. That's one reason I stopped using coveralls. :wink:

jfairley:
BTW, I don't know what you bower release process looks like, and I'm not familiar with "yuidoc", so let me know if I can update this to better support those libraries.

leandromoreira:
What do you think about this @towerz and @tjenkinson ?

towerz:
I would hold this change for a while. I already had a discussion with someone about this, I can't remember exactly who/when. The problem with publishing only the dist folder is that when someone uses webpack or babel in his project, we'll be forcing them to use the compiled version of clappr. Installing only the dist files is only recommended if you are publishing a library which should not be bundled into someone's code or have its inner workings exposed.

I've seen packages distributed just the way we do as well as packages that only include the compiled files in their distribution. Since the beginning we wanted the user to either build clappr into their own product as well as use clappr by itself with no changes. The way we distribute clappr files on npm supports both ways. One of the biggest problems this change will introduce is that the players that include clappr into their own compiled bundle may have some compilation issues introduced (for example, many objects that aren't exported but could be customized), as well as duplicating babel's ES6 runtime, and resuting in an unoptimized bundle, since the entire clappr bundle, with it's own UMD module exports will be bundled into another UMD module.

jfairley:
@towerz I can't say I'm familiar with the types of potential issues you sound familiar with. If you choose to not merge, I understand. 

I do hope you give some thought to #793, however. Your use of shrinkwrap forces me to have a preinstall script to remove it so I can build in linux. Alternatively, I know I can use node 5, but my dependency graph is all sorts of busted in npm 3, so that's not something I can easily jump into unfortunately.

leandromoreira:
hey @jfairley since we removed the shoryuken can we close this ?

jfairley:
https://github.com/clappr/clappr/pull/866 solves my immediate issue. Fine to close.

PR #823, opened 2016-02-03
Fixed bug where selected level id=0 would show AUTO

bikegriffith:
There was a null vs 0 discrepancy which caused the level selector plugin to exhibit some odd behavior.

_I need a little help here getting a test working.  I'm at the point where I'm monkeypatching hls.js, but that isn't where we want to be.  Any ideas for a more appropriate pattern? _


leandromoreira:
thank you so much @bikegriffith

bikegriffith:
@leandromoreira Certainly! However, you might want to revert that commit because I committed with a broken test.  I can undo the test entirely or hack away at it more to get it working.

bikegriffith:
I updated my original source branch with a new commit to fix that test.

PR #825, opened 2016-02-04
Added "getStartTimeOffset()"

tjenkinson:
This returns the time that "0" now represents relative to when playback started.

Useful for plugins that want to place items in the player which correspond with a particular point in time when it's a live stream with a sliding window. The plugins will need to know how the media sliding.

It cannot be assumed that the media slides at a constant rate. It actually jumps because it maps to chunks, and the server may decide not to start removing chunks from the start of the stream until it's reached a certain duration.

Ref: https://github.com/tjenkinson/clappr-markers-plugin/issues/3

### DIFF
--- a/src/base/playback.js
+++ b/src/base/playback.js
@@ -70,10 +70,20 @@ export default class Playback extends UIObject {
    */
   seekPercentage(percentage) {/*jshint unused:false*/}
 
+
+  /**
+   * The time that "0" now represents relative to when playback started.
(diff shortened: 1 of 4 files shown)


leandromoreira:
:+1: I'm up for this too should we implement this on [FlasHLS](https://github.com/clappr/clappr/blob/master/src/playbacks/flashls/flashls.js) too?

PR #829, opened 2016-02-08
HLS playback duration now excludes content past the live sync point

tjenkinson:
Users should not seek past the live sync point because otherwise their buffer might run out before the latest chunks arrive. Also seeking to the end of the stream, ie the getDuration() value, would sometimes go wrong and end up returning to the beginning. This seems to fix that.


leandromoreira:
as always thank you :smile: @tjenkinson

PR #832, opened 2016-02-10
Updated Babel to 6.x

jlmcdonald:
When updating Babel (so as to be able to build hls.js 0.5.x), it's also necessary to include babel-plugin-add-module-exports and to modify the webpack build process to work with new version of Babel.


webpack-base-config.js
```diff
@@ -11,7 +11,8 @@ module.exports = {
     loaders: [
       {
         test: /\.js$/,
-        loader: 'babel',
+        loader: 'babel-loader',
+	query: { presets: ['es2015'] }
```
tjenkinson: Is this needed? Isn't the Babel.rc file enough?

PR #836, opened 2016-02-10
Hide duration (current time) if using actual time in seek time component

tjenkinson:
When using the actualLiveTime feature on a live stream, currently the duration will be shown as the current time. This removes this as I don't think it's really useful.

@cowai you happy with this change? If you still want to be able to show the actual time as the duration it could be changed to be another config option instead.


leandromoreira:
once again tks

PR #861, opened 2016-03-01
enable some events to html image playback

avreg:


leandromoreira:
Thank you so much @avreg

PR #865, opened 2016-03-02
Fix build process to allow hls.js webworkify to function inside clappr webpack build

bikegriffith:
Per https://github.com/dailymotion/hls.js/issues/187, the hls.js webworker can't function after being bundled by webpack so demuxing is always performed inline on clappr.

This pull request uses webworkify-webpack and the webpack NormalModuleReplacement plugin to fix the bundle so the demuxing can be performed in the web worker.


leandromoreira:
Thanks @bikegriffith can you remove the dist from your commit before I accept it?

bikegriffith:
@leandromoreira Removed.

leandromoreira:
Thanks @bikegriffith

Issue #859, opened 2016-02-28
Player doesn't load in IE10


philcrump:
**Browser**: IE10

**OS**: Windows 7

**Clappr Version**: 0.2.35

**Steps to reproduce**: 

* Open http://cdn.clappr.io/ in IE10
* Click 'Run'
* Player doesn't load, error message: "Unable to get property 'html' of undefined or null reference"

**Did you try to reproduce this issue at http://cdn.clappr.io/**: Yes

Error is on line 8685 of http://cdn.clappr.io/latest/clappr.js:

> this.$el.html(this.template({ settings: this.settings }));

"SCRIPT5007: Unable to get property 'html' of undefined or null reference 
clappr.js, line 8685 character 8"

Call stack (in IE format):

> 	render, clappr.js, line 8685
> 	createMediaControl, clappr.js, line 3953
> 	setupMediaControl, clappr.js, line 3941
> 	Core, clappr.js, line 3719
> 	create, clappr.js, line 3551
> 	attachTo, clappr.js, line 475
> 	Global code, cdn.clappr.io, line 99

Top code line (/: line 99) is:

> player.attachTo(playerElement);

philcrump:
No change (issue still present) with 0.2.36

bikegriffith:
The problem occurs when the MediaControl tries to render and calls `this.$el.html(this.template({ settings: this.settings }));` but `this.$el` is undefined.  

Perhaps there's some issue in IE10 with the chained `super` calls in the constructor of MediaControl, UIObject, BaseObject, ...

bikegriffith:
Looks like this might help: https://github.com/babel/babel/tree/master/packages/babel-plugin-transform-proto-to-assign

PR #867, opened 2016-03-03
Use loose class transformation in babel to support IE10

bikegriffith:
Fixes https://github.com/clappr/clappr/issues/859


leandromoreira:
Thanks again man :beers:

PR #873, opened 2016-03-07
Fix "back to live" in flasHLS playback

### LINKED ISSUE #837: DVR - Back to live issue with flash plugin on 0.2.34
I found an issue with clappr 0.2.34. 

I am generating a live HLS stream with NGINX, if I use HLS.JS the DVR works as expected:

1)Pause video
2)Click back to live
3)The "back to live" text dissapears and "live" is displayed again

However with the flash plugin the "back to live" text stays on screen and "live" does not come back.

Thanks,

Adam

tjenkinson:
Also remove `timeUpdate()` because it doesn't appear to be used, and changed live boundary time from 5 seconds to 3 seconds to match hlsjs playback.

Fixes #837

### DIFF
--- a/src/playbacks/flashls/flashls.js
+++ b/src/playbacks/flashls/flashls.js
@@ -476,7 +476,7 @@ export default class FlasHLS extends BaseFlashPlayback {
       } else {
         this.updateCurrentState(state)
         this.hasEnded = true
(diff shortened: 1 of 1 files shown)


leandromoreira:
Thanks :)

PR #874, opened 2016-03-07
Fix flasHLS playback so that LOADED_METADATA handlers called after playerPlay() call

tjenkinson:
Previously if a user called `seek()` in a `LOADED_METADATA` event handler `playerPlay()` would then be called straight after resulting in playback starting at 0.


leandromoreira:
you're on fire :beers:

PR #876, opened 2016-03-09
Run onLevelUpdated after playback type update

### LINKED ISSUE #870: UI glitches and weird behaviour at the end of a HLS live stream
**Browser**: Chrome Version 48.0.2564.116 (64-bit)

**OS**: OSX 10.11.3 (15D21)

**Clappr Version**: 0.2.38

**Steps to reproduce**:

1. Set the player source to be a live .m3u8 playlist file (live = that is currently being appended parts to).
2. Play the video till the end of the live stream.

I was expecting to see the same behaviour as reaching the end of an offline video but instead the progress timecode is sometimes going past the total duration timecode, the player turns black, with a white box jumping when hovering the mouse over the player, loading graphics to appear and having weird glitches.

I could reproduce it on http://cdn.clappr.io/

screen shot 2016-03-07 at 16 07 06

kolpax:
Fixes #870

### DIFF
--- a/src/playbacks/hls/hls.js
+++ b/src/playbacks/hls/hls.js
@@ -148,6 +148,7 @@ export default class HLS extends HTML5VideoPlayback {
   updatePlaybackType(evt, data) {
     this.playbackType = data.details.live ? Playback.LIVE : Playback.VOD
     this.fillLevels()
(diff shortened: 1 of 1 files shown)


leandromoreira:
Thanks @kolpax and @tjenkinson

PR #881, opened 2016-03-11
Send the entire level object

bikegriffith:
A recent commit https://github.com/clappr/clappr/commit/aa4effd3b524475d2a161ceca682ca9b105c68dc changed the label of the levels from the height (e.g. 720p) to the bitrate in kbps.  This affects level selector plugins.  We should allow the plugins to decide how to format their own labels by sending everything (including height and bitrate) to the PLAYBACK_LEVELS_AVAILABLE event.


flavioribeiro:
This makes sense to me!

PR #882, opened 2016-03-11
Flash parity with hls.js for level metadata

bikegriffith:
Following up on PR #881


bikegriffith:
Nevermind; I see you already did this.

leandromoreira:
:smiley:

PR #883, opened 2016-03-11
Allow options to disable the media control from opening when buffer starts

bikegriffith:
`options.mediacontrol.showOnBuffer`: default true

This is backward compatible with existing behavior.  You can set it to false to change behavior. We prefer to not have the scrub bar show up during small seek operations that trigger a buffer event, especially in full screen mode.  The buffer dots are enough to indicate a problem; then mouse movement can open the scrubber.


bikegriffith:
Don't merge this yet.  It's still not completely working.

PR #884, opened 2016-03-11
Avoid triggering extraneous MediaControl events which can cause playbar to show unexpectedly

bikegriffith:
The Poster Plugin was enabling/disabling the media control bar on every buffer, bufferfull, and play events.  This caused the media control bar to show up unexpectedly (for example during hls.js's instantaneous jumps over holes in a segment).  The container now only triggers the CONTAINER_MEDIACONTROL_ENABLE and CONTAINER_MEDIACONTROL_DISABLE events when they change.

### DIFF
--- a/src/components/container/container.js
+++ b/src/components/container/container.js
@@ -363,14 +363,14 @@ export default class Container extends UIObject {
   }
 
   disableMediaControl() {
-    if (this.mediaControlDisabled !== true) {
+    if (!this.mediaControlDisabled) {
       this.mediaControlDisabled = true
       this.trigger(Events.CONTAINER_MEDIACONTROL_DISABLE)
     }
   }
 
   enableMediaControl() {
-    if (this.mediaControlDisabled !== false) {
+    if (this.mediaControlDisabled) {
       this.mediaControlDisabled = false
       this.trigger(Events.CONTAINER_MEDIACONTROL_ENABLE)
     }
--- a/src/components/container/container.js
+++ b/src/components/container/container.js
@@ -363,13 +363,17 @@ export default class Container extends UIObject {
   }
 
   disableMediaControl() {
-    this.mediaControlDisabled = true
-    this.trigger(Events.CONTAINER_MEDIACONTROL_DISABLE)
+    if (this.mediaControlDisabled !== true) {
+      this.mediaControlDisabled = true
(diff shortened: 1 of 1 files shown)


src/components/container/container.js
```diff
@@ -363,13 +363,17 @@ export default class Container extends UIObject {
   }
 
   disableMediaControl() {
-    this.mediaControlDisabled = true
-    this.trigger(Events.CONTAINER_MEDIACONTROL_DISABLE)
+    if (this.mediaControlDisabled !== true) {
```
leandromoreira: I think we be positive abou this statement:
```javascript
// initialize this value at constructor (or check the best place)
this.mediaControlEnabled = false

//disableMediaControl() 
// ... easier to read
if (this.mediaControlEnabled) {
  this.mediaControlEnabled = false
  //...
}

//enableMediaControl() 
// ... easier to read
if (!this.mediaControlEnabled) {
  this.mediaControlEnabled = true
  //...
}
```

What do you think?

towerz: I agree with @leandromoreira. Using `mediaControlEnabled` seems easier to read, and his suggestion gets rid of the weird `!== true` comparison for a boolean value, since we would never have an `undefined` value.

bikegriffith:
Thanks for the feedback (that was me just being paranoid about truthiness).  Fixed to be more readable.

src/components/container/container.js
```diff
@@ -363,13 +363,17 @@ export default class Container extends UIObject {
...
+    if (!this.mediaControlDisabled) {
+      this.mediaControlDisabled = true
+      this.trigger(Events.CONTAINER_MEDIACONTROL_DISABLE)
+    }
   }
 
   enableMediaControl() {
-    this.mediaControlDisabled = false
-    this.trigger(Events.CONTAINER_MEDIACONTROL_ENABLE)
+    if (this.mediaControlDisabled) {
```
leandromoreira: Does it work for the first time when clappr will try to enable the mediaControl? (`undefined`)

bikegriffith: @leandromoreira It's never undefined ... it's explicitly initialized to false.

leandromoreira:
@clappr/core can we merge this? it is looking good.

Issue #887, opened 2016-03-12
hlsjs retry manifest get


ronag:
Not sure if this is in clappr or hlsjs. But if it takes the server a long time to respond to the manifest (m3u8) request then the request gets cancelled and is never re-tried.

I would like to automatically retry the requests after a timeout. Any advice on how to achieve this?

leandromoreira:
https://github.com/clappr/clappr/issues/813

ronag:
Closing in favour of PR.

Issue #813, opened 2016-01-29
Automatic HLS recover


cowai:
Hi,
There should be an automatic recover media option to Clappr.
There is too many things that can go wrong with hls.js and forcing a recover when the browser wont play a fragment would temporarily solve many issues until hls.js is 100% ready.

Any thoughts?

leandromoreira:
@cowai aren't we able to instrument that by only using its [config](https://github.com/dailymotion/hls.js/blob/master/API.md#fine-tuning) like `fragLoadingMaxRetry`?

cowai:
@leandromoreira 
On the demo page of hls.js there is a Auto Media Recover option which triggers hls.RecoverMediaError().
I dont think fragLoadingMaxRetry covers every corner case. RecoverMediaError() is kind of a last resort for all types of errors.

leandromoreira:
:) I get it! @cowai it makes sense

leandromoreira:
http://dailymotion.github.io/hls.js/demo/

me-vlad:
@cowai , @leandromoreira  good idea, I think it will be nice to get recovery option idependent from playback type (hlsjs, flashls, dash, etc)

shaharmor:
The way its done in hls.js is:

1. listen to the error even on the video tag:

```javascript
videotag.addEventListener('error', function(e) {
    var mediaError = e.currentTarget.error;

    if (mediaError.code === mediaError.MEDIA_ERR_DECODE) {
        recover();
    }
});
```

And also to hlsjs errors:

```javascript
hls.on(Hls.Events.ERROR, function(event, data) {
    if (data.fatal) {
        switch (data.type) {
            case Hls.ErrorTypes.NETWORK_ERROR:
                hls.startLoad();
                break;
            case Hls.ErrorTypes.MEDIA_ERROR:
                recover();
                break;
            default:
                // cannot recover
                break;
        }
    }
});
```

where the `recover()` function is:

```javascript
function recover() {
    if (!_recoveredDecodingError) {
        _recoveredDecodingError = true;
        hls.recoverMediaError();
    } else if (!_recoveredAudioCodecError) {
        _recoveredAudioCodecError = true;
        hls.swapAudioCodec();
        hls.recoverMediaError();
    } else {
        // failed recovery
    }
}
```

and `_recoveredDecodingError` & `_recoveredAudioCodecError` are instance members.

Any chance to get this implemented?

ronag:
[...]

PR #890, opened 2016-03-13
hls: Recover on network error.

ronag:
https://github.com/clappr/clappr/issues/813
https://github.com/clappr/clappr/issues/887

### DIFF
--- a/src/playbacks/hls/hls.js
+++ b/src/playbacks/hls/hls.js
@@ -147,8 +147,7 @@ export default class HLS extends HTML5VideoPlayback {
           this.recover()
           break
         default:
-          Log.error(`hlsjs: could not recover from error, evt ${evt}, data ${data} `)
-          this.trigger(Events.PLAYBACK_ERROR, "error description", this.name)
+          this.trigger(Events.PLAYBACK_ERROR, `hlsjs: could not recover from error, evt ${evt}, data ${data} `, this.name)
           break
       }
     }
--- a/src/playbacks/hls/hls.js
+++ b/src/playbacks/hls/hls.js
@@ -44,6 +44,7 @@ export default class HLS extends HTML5VideoPlayback {
     // be ignored. "playableRegionDuration" does not consider this
     this.playableRegionDuration = 0
     options.autoPlay && this.setupHls()
+    this.recoverAttemptsRemaining = options.hlsRecoverAttempts || 16
   }
 
   setupHls() {
@@ -134,7 +135,8 @@ export default class HLS extends HTML5VideoPlayback {
   }
 
   onError(evt, data) {
-    if (data && data.fatal) {
+    if (data && data.fatal && this.recoverAttemptsRemaining > 0) {
+      this.recoverAttemptsRemaining -= 1
       switch (data.type) {
         case Hls.ErrorTypes.NETWORK_ERROR:
           hls.startLoad()
--- a/src/playbacks/hls/hls.js
+++ b/src/playbacks/hls/hls.js
@@ -53,7 +53,21 @@ export default class HLS extends HTML5VideoPlayback {
     this.hls.on(HLSJS.Events.LEVEL_UPDATED, (evt, data) => this.onLevelUpdated(evt, data))
     this.hls.on(HLSJS.Events.LEVEL_SWITCH, (evt,data) => this.onLevelSwitch(evt, data))
     this.hls.on(HLSJS.Events.FRAG_LOADED, (evt, data) => this.onFragmentLoaded(evt, data))
+    this.hls.on(HLSJS.Events.ERROR, (evt, data) => this.onError(evt, data))
     this.hls.attachMedia(this.el)
+  }
+
+  recover() {
+    if (!_recoveredDecodingError) {
+      _recoveredDecodingError = true
+      hls.recoverMediaError()
+    } else if (!_recoveredAudioCodecError) {
+      _recoveredAudioCodecError = true
+      hls.swapAudioCodec()
+      hls.recoverMediaError()
+    } else {
+        // failed recovery
+    }
   }
 
   // override
@@ -117,6 +131,22 @@ export default class HLS extends HTML5VideoPlayback {
     }
     this.settings.seekEnabled = this.isSeekEnabled()
     this.trigger(Events.PLAYBACK_SETTINGSUPDATE)
+  }
+
+  onError(evt, data) {
+    if (data.fatal) {
+      switch (data.type) {
+        case Hls.ErrorTypes.NETWORK_ERROR:
+          hls.startLoad()
+          break
+        case Hls.ErrorTypes.MEDIA_ERROR:
+          recover()
+          break
+        default:
+          // cannot recover
+          break
+      }
+    }
   }
 
   onTimeUpdate() {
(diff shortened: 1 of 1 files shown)


ronag:
Somewhere we also have to reset `_recoveredDecodingError` and `_recoveredAudioCodecError`...

src/playbacks/hls/hls.js
```diff
@@ -119,6 +133,22 @@ export default class HLS extends HTML5VideoPlayback {
     this.trigger(Events.PLAYBACK_SETTINGSUPDATE)
   }
 
+  onError(evt, data) {
+    if (data && data.fatal) {
+      switch (data.type) {
+        case Hls.ErrorTypes.NETWORK_ERROR:
+          hls.startLoad()
```
tjenkinson: Should there maybe be some kind of delay before trying to load again? Don't want to start a dos attack on the server if it is returning an error :P

And maybe a limit on the number of attempts?

ronag: @tjenkinson: Isn't there a timeout on the request itself already? Yes, a retry limit might be a good idea.

tjenkinson: @ronag I'm not sure I haven't looked, but that is something that hlsjs has control over so if it changed in the future then it would be our fault. If we have a limit though then I think a few repeated attempts wouldn't really matter.

ronag: @tjenkinson: I'm unsure how to add a delay because things can happen between starting the delay and recovering which might cause the recover call to be invalid/bad.

tjenkinson: @ronag yes asynchronous things in javascript can get complicated quickly. I reckon just a limit would be good enough.

ronag: @tjenkinson: done

ronag: @tjenkinson: Though we should probably also reset the attempt counter when it successfully recovered...

src/playbacks/hls/hls.js
```diff
@@ -119,6 +134,23 @@ export default class HLS extends HTML5VideoPlayback {
...
+    if (data && data.fatal && this.recoverAttemptsRemaining > 0) {
+      this.recoverAttemptsRemaining -= 1
+      switch (data.type) {
+        case Hls.ErrorTypes.NETWORK_ERROR:
+          hls.startLoad()
+          break
+        case Hls.ErrorTypes.MEDIA_ERROR:
+          recover()
+          break
+        default:
```
leandromoreira: I think it's better to log this.
`Log.error(`hlsjs: this error could not be recovered, evt ${evt}, data ${data} `)`

leandromoreira:
@clappr/core we decided that at least a two persons should review a PR to then accept it. For me this is good

leandromoreira:
agreed @tjenkinson

src/playbacks/hls/hls.js
```diff
@@ -119,6 +134,25 @@ export default class HLS extends HTML5VideoPlayback {
...
+    if (data && data.fatal && this.recoverAttemptsRemaining > 0) {
+      this.recoverAttemptsRemaining -= 1
+      switch (data.type) {
+        case Hls.ErrorTypes.NETWORK_ERROR:
+          Log.warning(`hlsjs: trying to recover from network error, evt ${evt}, data ${data} `)
+          hls.startLoad()
+          break
+        case Hls.ErrorTypes.MEDIA_ERROR:
+          Log.warning(`hlsjs: trying to recover from media error, evt ${evt}, data ${data} `)
+          recover()
```
leandromoreira: I think you need to call `this.recover()`

src/playbacks/hls/hls.js
```diff
@@ -53,9 +54,23 @@ export default class HLS extends HTML5VideoPlayback {
...
     this.hls.on(HLSJS.Events.LEVEL_SWITCH, (evt,data) => this.onLevelSwitch(evt, data))
     this.hls.on(HLSJS.Events.FRAG_LOADED, (evt, data) => this.onFragmentLoaded(evt, data))
+    this.hls.on(HLSJS.Events.ERROR, (evt, data) => this.onError(evt, data))
     this.hls.attachMedia(this.el)
   }
 
+  recover() {
+    if (!this._recoveredDecodingError) {
+      this._recoveredDecodingError = true
+      hls.recoverMediaError()
```
leandromoreira: `this.hls` instead of `hls`

src/playbacks/hls/hls.js
```diff
@@ -119,6 +134,25 @@ export default class HLS extends HTML5VideoPlayback {
...
+          this.hls.startLoad()
+          break
+        case HLSJS.ErrorTypes.MEDIA_ERROR:
+          Log.warning(`hlsjs: trying to recover from media error, evt ${evt}, data ${data} `)
+          this.recover()
+          break
+        default:
+          Log.error(`hlsjs: could not recover from error, evt ${evt}, data ${data} `)
+          break
+      }
```
bikegriffith: How does this affect the current triggering of `Events.PLAYBACK_ERROR`?  (I don't see when we're explicitly triggering it presently, but I know I've seen it before).

leandromoreira: You're probably right, we need to trigger PLAYBACK_ERROR for default and the failed recover, good catch.

src/playbacks/hls/hls.js
```diff
@@ -53,9 +54,23 @@ export default class HLS extends HTML5VideoPlayback {
...
+  recover() {
+    if (!this._recoveredDecodingError) {
+      this._recoveredDecodingError = true
+      this.hls.recoverMediaError()
+    } else if (!this._recoveredAudioCodecError) {
+      this._recoveredAudioCodecError = true
+      this.hls.swapAudioCodec()
+      this.hls.recoverMediaError()
+    } else {
+        // failed recovery
```
leandromoreira: We should log this failed try and also trigger a playback_error

src/playbacks/hls/hls.js
```diff
@@ -119,6 +134,25 @@ export default class HLS extends HTML5VideoPlayback {
...
+        case HLSJS.ErrorTypes.NETWORK_ERROR:
+          Log.warning(`hlsjs: trying to recover from network error, evt ${evt}, data ${data} `)
+          this.hls.startLoad()
+          break
+        case HLSJS.ErrorTypes.MEDIA_ERROR:
+          Log.warning(`hlsjs: trying to recover from media error, evt ${evt}, data ${data} `)
+          this.recover()
+          break
+        default:
+          Log.error(`hlsjs: could not recover from error, evt ${evt}, data ${data} `)
```
leandromoreira: We should also trigger a playback_error

ronag: @leandromoreira: Not sure what/how that is.

leandromoreira: @ronag no problem, usually when we face an error that we can't do nothing we trigger a playback error therefore the player can show something for the final user. 

The [tag video](https://github.com/clappr/clappr/blob/master/src/playbacks/html5_video/html5_video.js#L291), for instance, it handles the error this way:
```javascript
  onError(event) {
    this.trigger(Events.PLAYBACK_ERROR, this.el.error, this.name)
  }
```
In our case we can trigger an even like this but changing the reason :)

src/playbacks/hls/hls.js
ronag: Please show how that should look...

leandromoreira: you should trigger in each point of no turning back (failed recovery, `default`)
```javascript
this.trigger(Events.PLAYBACK_ERROR, "error description", this.name)
```

leandromoreira:
Thanks @ronag

Issue #888, opened 2016-03-13
img webp support


ronag:
I see no reason why img playback shouldn't work with webp?

ronag:
Closing in favour of PR.

PR #891, opened 2016-03-13
img_html: Support webp

ronag:
https://github.com/clappr/clappr/issues/888


leandromoreira:
@clappr/core we decided that at least a two persons should review a PR to then accept it. For me this is good

leandromoreira:
@clappr/core can we merge this?

PR #906, opened 2016-03-24
core: fix mimetype when setting it using configure()

jussike:
When setting source using configure(), mimetype was null. If source had not appropriate file extension, source could not be played.

### DIFF
--- a/src/components/core/core.js
+++ b/src/components/core/core.js
@@ -301,7 +301,7 @@ export default class Core extends UIObject {
     var sources = options.source || options.sources
 
     if (sources) {
-      this.load(sources)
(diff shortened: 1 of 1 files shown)


src/components/core/core.js
```diff
@@ -301,7 +301,7 @@ export default class Core extends UIObject {
     var sources = options.source || options.sources
 
     if (sources) {
-      this.load(sources)
+      this.load(sources, this.options.mimeType)
```
leandromoreira: Thanks @jussike :) can you make it 
```javascript
options.mimeType || this.options.mimeType
```

jussike:
I changed it as you suggested.

leandromoreira:
Thank you so much @jussike

PR #908, opened 2016-03-28
Added "npm update" to bump script

tjenkinson:
To make sure that new versions always have the latest specified versions of the dependencies at the time of release.

### DIFF
--- a/bump
+++ b/bump
@@ -1,5 +1,10 @@
 #!/bin/sh
 
+update_dependencies() {
+  echo 'updating dependencies'
+  npm update > /dev/null
+}
+
 update_version() {
   current_tag=$(git describe --abbrev=0 --tags)
   echo 'bump from '$current_tag' to '$1
@@ -55,6 +60,7 @@ Your happy butler.\n
 }
 
 main() {
+  update_dependencies
   update_version $1
   update_bower
   build
--- a/bump
+++ b/bump
@@ -18,9 +18,9 @@ update_bower() {
 
 build() {
   echo 'building clappr.js'
-  npm run build > /dev/null
+  npm run build --silent
   echo 'building clappr.min.js'
-  npm run release > /dev/null
+  npm run release --silent
 }
 
 run_tests() {
--- a/bump
+++ b/bump
@@ -2,7 +2,7 @@
 
 update_dependencies() {
   echo 'updating dependencies'
-  npm update > /dev/null
+  npm update --silent
 }
 
 update_version() {
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -11,7 +11,7 @@ if (process.env.npm_lifecycle_event === 'release') {
     output: {comments: false}
   }));
 } else {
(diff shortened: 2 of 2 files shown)


flavioribeiro:
LGTM

towerz:
@tjenkinson why don't we replace the redirect to /dev/null with the --silent flag? This way, messages that npm consider that should appear would be printed anyway.

towerz:
@tjenkinson perhaps it would be better to keep all npm calls with silent, yes.

leandromoreira:
[joke]let's make npm silent once for all[/joke]

towerz:
@tjenkinson it does exist and actually works on update/run. When using on npm run, it only removes the header of npm log:
```
npm run build --silent
clean-webpack-plugin: /Users/bruno/workspace/player/clappr/dist has been removed.
Hash: c6987ef2650db63ae438
Version: webpack 1.12.14
Time: 12790ms
                               Asset       Size  Chunks             Chunk Names
a8c874b93b3d848f39a71260c57e3863.cur  326 bytes          [emitted]
4b76590b32dab62bc95c1b7951efae78.swf     3.2 kB          [emitted]
d218edf766218c19b416107bfb05ef0f.swf      63 kB          [emitted]
38861cba61c66739c1452c3a71e39852.ttf    32.7 kB          [emitted]
                           clappr.js     888 kB       0  [emitted]  main
    + 155 hidden modules
```
Without `--silent`:
```
npm run build

> clappr@0.2.40 build /Users/bruno/workspace/player/clappr
> webpack --progress

clean-webpack-plugin: /Users/bruno/workspace/player/clappr/dist has been removed.
Hash: c6987ef2650db63ae438
Version: webpack 1.12.14
Time: 13549ms
                               Asset       Size  Chunks             Chunk Names
a8c874b93b3d848f39a71260c57e3863.cur  326 bytes          [emitted]
4b76590b32dab62bc95c1b7951efae78.swf     3.2 kB          [emitted]
d218edf766218c19b416107bfb05ef0f.swf      63 kB          [emitted]
38861cba61c66739c1452c3a71e39852.ttf    32.7 kB          [emitted]
                           clappr.js     888 kB       0  [emitted]  main
    + 155 hidden modules
```

towerz:
This extra output seems to come from webpack itself, so we could configure it to be less verbose when building. There seems to be some configs to make webpack less verbose.

towerz:
@tjenkinson this message is from the clean-webpack-plugin. I just found that there's an option `verbose: false` which will prevent this message from being printed. Try changing this line on webpack.config.js:
```javascript
webpackConfig.plugins.push(new Clean(['dist'], {verbose: false}));
```

towerz:
I wouldn't redirect 1>/dev/null

towerz:
@tjenkinson it would be better, but it's not actually mandatory not to print anything. As I mentioned, this was made to keep things easier to read. I would rather let it print the chunks and avoid redirecting any output to /dev/null.

towerz:
It is, thanks!

PR #910, opened 2016-03-29
Make bump script fail on more errors than "npm test"

tjenkinson:
So that if dependencies fail to update, or there's an issue updating the package files, a push and publish can't happen.

Also specifically adds "package.json" and "bower.json" to git commit instead of everything.

First time I've really messed around with bash scripting so there might be issues. Also I haven't run it because on windows, and don't want to accidentally release something :P

### DIFF
--- a/bump
+++ b/bump
@@ -1,14 +1,14 @@
-#!/bin/sh
+#!/bin/bash
 
 update_dependencies() {
-  echo 'updating dependencies'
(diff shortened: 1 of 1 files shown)


leandromoreira:
@tjenkinson is the new bash master ! it's good

towerz:
Great, merged.

PR #914, opened 2016-03-31
git add the dist folder in the bump script

tjenkinson:
This is why the dist folder still has the builds from the last version :/

PR #915, opened 2016-03-31
Update dist to 0.2.41

tjenkinson:
Previously incorrect because of https://github.com/clappr/clappr/pull/914


leandromoreira:
thanks :)

PR #916, opened 2016-03-31
Fix media control button alignment

tjenkinson:
Seemed to have an issue on a webpage with bootstrap for some reason, and this fixed it.

I don't think the font anti aliasing should be there anymore given there's not text, but can put it back in if that still does magic with svgs.

Tested in chrome, firefox, internet explorer and edge.

Before:
image

After:
image


towerz:
Thanks, @tjenkinson. You have spared me of this task.

About the font smoothing, it has no impact when used on svgs directly.

PR #929, opened 2016-04-05
Ensure dvr only enables when duration has stabilized

tjenkinson:
The duration will initially be the actual duration until the real duration is longer the hlsjs's live sync point, then it drops back down to hide the content after the sync point (to prevent someone seeking to
a point which will cause their buffer to run out before the next chunk arrives).

This ensures that the dvr can only be enabled when the duration is not including the content after the live sync point.

Feels a little bit hacky with the duration being reduced when there is enough content, so if you can think of a better way let me know. I think the nicest would be only making the playback ready when the actual duration goes past the sync point, but this means when a live stream starts the user would be shown the buffering indicator for a while, unless the live stream was only made available to the player after there was already enough chunks in the playlist to bring the duration past the live sync point already.


towerz:
Just tested here, great job @tjenkinson. Thanks!

PR #931, opened 2016-04-06
Hls: onLevelUpdated(): Update duration of mediasource. Fixes #917.

### LINKED ISSUE #917: HLS: seeking is broken after changing live mode to vod on the fly
***1 Upvote*** Environment:
FF 45, Chrome 49.0.2623.87, others. **Only Safari works as expected.**
Linux, Windows, OS X
Latest Clappr

Steps to reproduce:
* Provide a HLS-livestream (without #EXT-X-ENDLIST)
* Start playback
* Update your livestream to VOD (#EXT-X-ENDLIST as last line)
* Wait until Clappr detect VOD mode
* Try to seek near of end

What happened:
* The player allows you to try to seek but seek bar will slide back to near of original place

Expected behaviour:
* HLS VOD stream should be seekable

Note:
* After refreshing, the error is gone because now the stream is vod all the time.

jussike:
This fixes https://github.com/clappr/clappr/issues/917 for me.
Is there any possible side effect?

### DIFF
--- a/src/playbacks/hls/hls.js
+++ b/src/playbacks/hls/hls.js
@@ -215,7 +215,7 @@ export default class HLS extends HTML5VideoPlayback {
       }
     }
     if (newDuration !== this.playableRegionDuration) {
-      this.playableRegionDuration = newDuration
+      this.hls.bufferController.mediaSource.duration = this.playableRegionDuration = newDuration
       this.onDurationChange()
     }
   }


towerz:
I agree with @tjenkinson. It's probably a bug on hls.js and should be fixed there instead. Since #917 is closed, I believe @jussike agrees with us.

We'll close this PR, but if someone find an issue in Clappr that's related to this we can revisit this solution.

PR #938, opened 2016-04-11
Report correct times in PLAYBACK_PROGRESS event when hls with dvr

tjenkinson:
Previously the times reported in the event wouldn't take into consideration the media that has slid off the start of the timeline, in the case of a hls stream with sliding window dvr.


towerz:
Great job, @tjenkinson, thanks!

PR #940, opened 2016-04-12
flashls.js: Listen endlist-event and update playback type.

jussike:
Update playback type live->vod when endlist-event triggered.
See: https://github.com/mangui/flashls/pull/510


towerz:
Thanks @jussike, good one.

PR #942, opened 2016-04-13
trigger Events.PLAYBACK_ERROR on networkError

PNixx:


flavioribeiro:
Thanks @PNixx!

leandromoreira:
Thanks @PNixx and @flavioribeiro I think that the best way to fix this issue is triggering `PLAYBACK_ERROR` only after reaching the max numbers of attempt. With the solution we'd drop the recovering tries.


Do you agree?

```javascript
  onError(evt, data) {
    if (data && data.fatal && this.recoverAttemptsRemaining > 0) {
      this.recoverAttemptsRemaining -= 1
      switch (data.type) {
        case HLSJS.ErrorTypes.NETWORK_ERROR:
          Log.warn(`hlsjs: trying to recover from network error, evt ${evt}, data ${data} `)
          this.hls.startLoad()
          break
        case HLSJS.ErrorTypes.MEDIA_ERROR:
          Log.warn(`hlsjs: trying to recover from media error, evt ${evt}, data ${data} `)
          this.recover()
          break
        default:
          Log.error(`hlsjs: trying to recover from media error, evt ${evt}, data ${data} `)
          this.trigger(Events.PLAYBACK_ERROR, `hlsjs: could not recover from error, evt ${evt}, data ${data} `, this.name)
          break
      }
    } else {
        this.trigger(Events.PLAYBACK_ERROR, `hlsjs: could not recover from error after maximum number of attempts, evt ${evt}, data ${data} `, this.name)
    }
  }
```

leandromoreira:
We're going to release a version of Clappr therefore we'll need to rollback this commit, I'm sorry. :( But on  the other hand we're going to fix the issue :)

towerz:
I agree with @leandromoreira, triggering an error event should only occur when there's no more recovery attempts left.

PR #947, opened 2016-04-14
Don't trigger PLAYBACK_ERROR on non-fatal hlsjs errors

tjenkinson:
I don't think we should trigger a playback error if hlsjs is trying to fix it itself. Hlsjs will probably fire a fatal error if it tries to fix it and fails.

Also change `onError()` to `onHLSJSError()` so not to override `onError()` in `Html5Video`, and guarantee that `evt` and `data` will be from hlsjs.

This puts the functionality back to when playback error events were thrown in the last release.


towerz:
You're right, @tjenkinson. Our solution didn't take non-fatal errors into account. Thanks!

PR #966, opened 2016-04-29
Several enhancements in Player class

### LINKED ISSUE #677: player.setVolume() does not updates the localStorage value
On http://cdn.clappr.io/ website in console enter
```js
localStorage
```
This will return the
```json
{clappr.cdn.clappr.io.volume: "0"}
```
then enter
```js
player.setVolume(50);
```
and then again
```js
localStorage;
```
but if the volume is changed by pressing on volume control bar then localStorage returns
```json
{clappr.cdn.clappr.io.volume: "56.02678571428571"}
```

gamtiq:
I have made the following improvements in Player class:
* refactored `setVolume` and `getVolume` methods (this change should fix #677)
* changed implementation of `mute` and `unmute` methods to restore volume level
* added `start` method
* added `return this;` into interface methods so that make them chainable

### DIFF
--- a/src/components/player.js
+++ b/src/components/player.js
@@ -343,13 +343,17 @@ export default class Player extends BaseObject {
   /**
    * loads a new source.
    * @method load
-   * @param {Object} sources source or sources of video.
-   * sources can be a string or {source: <<source URL>>, mimeType: <<source mime type>>}
-   * @param {Object} mimeType a mime type, example: `'application/vnd.apple.mpegurl'`
+   * @param {Array|String} sources source or sources of video.
+   * An array item can be a string or {source: <<source URL>>, mimeType: <<source mime type>>}
+   * @param {String} mimeType a mime type, example: `'application/vnd.apple.mpegurl'`
+   * @param {Boolean} [startPlay=false] whether playing should be started immediately
    * @return {Player} itself
    */
-  load(sources, mimeType) {
-    this.core.load(sources, mimeType)
+  load(sources, mimeType, startPlay) {
+    this.core.load(sources, mimeType);
+    if (startPlay) {
+      this.play();
+    }
     return this;
   }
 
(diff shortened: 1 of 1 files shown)


leandromoreira:
Thank you so much @gamtiq 

@clappr/core any thoughts? 

- some fixes (great)
- returning player itself (I'm okay with this, we already adopted that for render)
- adding a new start method (this is where I need to think better, it is not feasible to mobile, at least in the first play)

towerz:
@leandromoreira I agree with you. Perhaps the `start` method should be analyzed more carefully.

We should consider moving these changes into the configure and load methods, instead of creating a new one that decides either reconfiguring or loading a new source based on the type of its parameters.

gamtiq:
I have added `start` method for convenience so that to make just one call instead of two (`load`/`configure` and `play`) when starting playing of new video.
Maybe it is worth not to add new method but change implementation of `play` method so that it would be possible to pass optional parameters specifying source that should be played.
If that is accepted I can delete `start` method and move the appropriate logic into `play` method.

towerz:
@gamtiq I was actually more inclined to add these changes you made on `start` to the `load`/`configure` methods. I understand you intended to automatically start the video on load when calling the `start` method.

Instead of doing that, I suggest to use the `autoPlay` semantics when calling `load`/`configure`. On `load`, we could add an `autoPlay` parameter to automatically start the video, and when `configure` is called, setting the `autoPlay` flag to true, the video would start, just like if you have called `play` afterwards.

gamtiq:
I agree with you. Indeed adding auto play ability for `load` method would be enough.
Should I make any additional changes in my pull request? Or do you plan to merge only necessary commits (excepting c594318)?

towerz:
It's up to you. Do you prefer implementing the additional changes yourself and adding to the PR or would you like us to merge the other commits except c594318?

gamtiq:
I can make additional changes in PR. If I correctly understand you, it is necessary only to add the third parameter for `load` method and to modify the method slightly. I will do it.

towerz:
Great job, thanks @gamtiq! 🍻

PR #975, opened 2016-05-07
Update compass-mixins to a fork with "compact" function fix

tjenkinson:
This [fork](https://github.com/tjenkinson/compass-mixins) of [compass-mixins](https://github.com/Igosuki/compass-mixins) includes [this fix](https://github.com/Igosuki/compass-mixins/pull/86). It looks like the official compass mixins repo might not be being maintained at the moment, but this could easily be reverted if the fix is merged into the official compass mixins repo.

With this fix clappr will build fine with the latest version of node-sass, so once this is merged https://github.com/clappr/clappr/pull/970 should be good to go as well.

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,7 +35,7 @@
     "check-dependencies": "^0.12.0",
     "clappr-zepto": "0.0.4",
     "clean-webpack-plugin": "^0.1.3",
-    "compass-mixins": "^0.12.7",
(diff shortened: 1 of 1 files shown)


leandromoreira:
@tjenkinson I think @bernardocamilo is also working on a solution for this.

towerz:
As we all discussed, we decided to use the solution proposed in #989, instead of keeping a fork of compass-mixins. Closing this PR. 

Thanks, @tjenkinson!

Issue #954, opened 2016-04-19
IE11 on Windows 10 change playback url using javascript not working


ibreakstuff93:
**Browser**:  IE 11.162.10586.0

*OS**: Windows 10

**Clappr Version**: 0.2.47

**Steps to reproduce**:

Step 1: Create javascript function as follows:
```
function changeurl(url){
player.stop();
player.load(url);
player.play();
}
```
Step 2: call javascript function to change url:
When a user clicks a button on the page, the javascript is executed with the url present, e.g:
```
onclick="changeurl(http://example.com/index.m3u8)"
```
This works in all browsers (including on iOS and Android (well done guys! :+1: ) but sadly not IE11.

The console in IE11 logs the following message after a user clicks:

![image](https://cloud.githubusercontent.com/assets/16917874/14637590/3dac3b0e-062a-11e6-871a-1404be0ae7ac.png)

tjenkinson:
Confirmed. Great username by the way you've lived up to it :P

ibreakstuff93:
Thanks - i do try :) Any ideas how long it might take to resolve?

ibreakstuff93:
Can i just check i've understood correctly, we can only use the stop() and pause() events when using html5 playback, load() isn't supported?

tjenkinson:
@ibreakstuff93 `load()` is still supported and should work.

ibreakstuff93:
@tjenkinson thank you for clarifying! I've just tried it again, but i'm still getting an error message in IE11. It's moved a few lines however this time...

![image](https://cloud.githubusercontent.com/assets/16917874/15112861/d470e0ee-15e8-11e6-8707-e9913b1755e7.png)

tjenkinson:
@ibreakstuff93 have you built master containing the fix?

Try this: https://cdn.rawgit.com/tjenkinson/215c450e27f2dc99efaf3d9e53fe1406/raw/756fdbf6f06744714a03456a08926db502a20fb7/clappr.js

ibreakstuff93:
![image](https://cloud.githubusercontent.com/assets/16917874/15113089/19eb7c8c-15ea-11e6-8419-664eafc7bfcd.png)

Thanks working now :+1:

ibreakstuff93:
Sorry for my stupid question, but will this fix get rolled up into the next release? (sorry, i'm still a newbie to github)

[...]

Issue #826, opened 2016-02-04
Firefox raises InvalidStateError when trying to destroy player


bikegriffith:
See jsfiddle here:
https://jsfiddle.net/d4hwuzwm/
(this loads a video, waits for it to start playing, then calls `destroy` 5 seconds later)

Calling `destroy` on a player instance in Firefox raises `InvalidStateError: An attempt was made to use an object that is not, or is no longer, usable`

This works fine in Chrome/Safari.

bikegriffith:
It looks like if you wait for the playback to stop, it won't raise the exception, but unfortunately, I don't have that luxury because I need to destroy the player immediately.

```
      console.log('Attempting to destroy player...');
      player.stop();
      player.core.getCurrentPlayback().once('playback:stop', function() {
 	     player.destroy();
      });
```

flavioribeiro:
reproduced here

PR #976, opened 2016-05-08
Only invoke stop() and pause() on html5video from hls playback if hls is initialised

tjenkinson:
In ie11 trying to set the 'currentTime' property to a value when hlsjs hasn't initialised throws "InvalidStateError".

Fixes https://github.com/clappr/clappr/issues/954
Fixes https://github.com/clappr/clappr/issues/826


leandromoreira:
Thanks once agin

PR #977, opened 2016-05-08
Reduce the frame rate of noop noise

### LINKED ISSUE #877: Infinite loop when src is empty or is not provided on load.
**Browser**: Chrome Version 49.0.2623.87 (64-bit)

**OS**: Ubuntu 14.04

**Clappr Version**: 0.2.38

**Steps to reproduce**:

* Open demo clappr page in text editor and erase the `src` or just leave empty quotes.
* Open demo page in the browser
* Check the CPU in task manager (the `loop()` function starts infinitely).

Actually there is a second problem -- intense error function, the `noise()` shouldn't be that intense in my opinion (comment `noise()` invocation and the cpu load would be about 3% vs 50-70%). I wish I could fix this, but I can't find the `NoOp()` trigger, so I don't know where to insert the `empty or poor source` check.

tjenkinson:
To reduce the cpu load.

Fixes #877

### DIFF
--- a/src/playbacks/no_op/no_op.js
+++ b/src/playbacks/no_op/no_op.js
@@ -28,6 +28,7 @@ export default class NoOp extends Playback {
   constructor(options) {
     super(options)
     this.options = options
+    this._noiseFrameNum = -1
   }
 
   render() {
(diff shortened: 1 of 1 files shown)


leandromoreira:
Thanks @coveralls lol

PR #999, opened 2016-05-21
Allow source to be a an instance of a class

tjenkinson:
useful for cases like https://github.com/tjenkinson/clappr-videocontext-playback

### DIFF
--- a/src/components/container_factory/container_factory.js
+++ b/src/components/container_factory/container_factory.js
@@ -38,14 +38,17 @@ export default class ContainerFactory extends BaseObject {
     if (typeof source === "string" || source instanceof String) {
       resolvedSource = source.toString()
     }
-    else {
+    else if (typeof source === "object" && !source.constructor) {
       resolvedSource = source.source.toString()
       if (source.mimeType) {
         mimeType = source.mimeType
       }
     }
+    else {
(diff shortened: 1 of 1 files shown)


towerz:
LGTM

thiagopnts:
wouldn't be easier to pass the instance as a parameter for the new playback? seems weird to have the core source support other things than strings

leandromoreira:
@thiagopnts do you mean?

```javascript
var _player = new Clappr.Player({sourceSolver: MyInstance})
```

thiagopnts:
your canplay method can use information from the source string to decide rather it should play or not, e.g. in the 360 plugin I just pass `video.mp4#360` as a source and my `canPlay` just match for `/#360$/`.

thiagopnts:
@tjenkinson would it work with something like this?

```javascript
constructor(options) {
    super(options)
    this._context = options.context
  //etc
 }
VideoContextPlayback.canPlay = (resource) => {
  return resource === 'video-context-playback'
}
```

thiagopnts:
aren't params from the player constructor passed to the other components anymore?

thiagopnts:
agree, I would just avoid adding code for edge cases that can be worked around using the plugin system

PR #1001, opened 2016-05-23
Revert "Allow source to be a an instance of a class"

tjenkinson:
It thinks that {} is a class instance.

This reverts commit 3c33821c8bfffd7f2a2052ad882cc530b3c23821.

This looks like it might work but it seems like a bit of a hack.
https://github.com/jonschlinkert/is-plain-object


leandromoreira:
Thanks @tjenkinson

Issue #998, opened 2016-05-20
Indicator is still showing if audioOnly is true


waster:
**Browser**: FF 46
**OS**: Ubuntu 16.04
**Clappr Version**: 0.2.51

Here is the code I tested on http://cdn.clappr.io:

```
var playerElement = document.getElementById("player-wrapper");

var player = new Clappr.Player({
  source: 'http://clappr.io/highline.mp4',
  baseUrl: '/latest',
  poster: 'http://clappr.io/poster.png',
  height: 360,
  width: 640,
  audioOnly:true
});

player.attachTo(playerElement);

var variant_streams = {
  'test': [
    {
      source: 'http://<URL>/playlist.m3u8',
      mimeType: 'application/x-mpegURL'
    }
  ]
};

player.load(variant_streams['test']);
player.play();

```
Source is the URL to the variant playlist of HLS with ts-segments with audio only stream.

I see that play button indicator is still showing while playing.

waster:
I also tested with the following multiple streams (with and without 'audioOnly:true'):

```
var variant_streams = {
  'test': [
    {
      source: 'http://<ULR_VARIANT1>/playlist.m3u8',
      mimeType: 'application/x-mpegURL'
    },
    {
      source: 'http://<ULR_VARIANT2>/playlist.m3u8',
      mimeType: 'application/x-mpegURL'
    },
    {
      source: 'http://<ULR_MP3>/stream.mp3',
      mimeType: 'audio/mp3'
    }
  ]
};

```
In this case bubbles spinner is showing infinitely while stream is playing.

leandromoreira:
Hey @waster I tried this code bellow at cdn and it works just fine, what is the problem are you facing?

```javascript
var playerElement = document.getElementById("player-wrapper");

var player = new Clappr.Player({
  source: 'http://clappr.io/highline.mp4',
  baseUrl: '/latest',
  poster: 'http://clappr.io/poster.png',
  height: 360,
  width: 640,
  audioOnly:true
});

player.attachTo(playerElement);

var variant_streams = {
  'test': [
    {
      source: 'http://www.streambox.fr/playlists/x36xhzz/x36xhzz.m3u8',
      mimeType: 'application/x-mpegURL'
    }
  ]
};

player.load(variant_streams['test']);
//player.play(); this will raise an error cause https://bugs.chromium.org/p/chromium/issues/detail?id=593273
```

leandromoreira:
[...]

PR #1004, opened 2016-05-26
Fix poster play so hides on click or play event only

tjenkinson:
or whilst buffering is happening (because buffering indicator will be in the same place)

Refs #998


leandromoreira:
LGTM

towerz:
🍻 @tjenkinson

PR #1005, opened 2016-05-26
fix bug if calling load() with a single source

tjenkinson:
source could be `{source, mimeType}`


leandromoreira:
LGTM

towerz:
Merged, thanks @tjenkinson

PR #1006, opened 2016-05-26
Fix 'sources' plugin so it still handles future sources

tjenkinson:
e.g if someone calls `load()` with an array

refs #998


towerz:
Nice fix

PR #1007, opened 2016-05-26
Add html5 tag video crossorigin

leandromoreira:


src/base/utils.js
```diff
@@ -216,6 +216,15 @@ export function getBrowserLanguage() {
   return null
 }
 
+export function setupObject(object, att) {
```
tjenkinson: I think you could use
```javascript
$.extend(object, att);
```
instead of this

src/playbacks/html5_video/html5_video.js
```diff
@@ -84,16 +84,19 @@ export default class HTML5Video extends Playback {
...
-      this.el.preload = options.preload ? options.preload: 'metadata'
-    }
+
+    var playbackConfig = (options.playbackConfig || {})
+    var preload = playbackConfig.preload || ((Browser.isSafari)?'auto':options.preload)
+
+    setupObject(this.el, {
+      loop: playbackConfig.loop || options.loop,
+      autoplay: playbackConfig.autoplay || options.autoPlay,
+      poster: playbackConfig.poster || options.poster,
```
tjenkinson: This property doesn't exist:
https://developer.mozilla.org/en/docs/Web/API/HTMLMediaElement

src/playbacks/html5_video/html5_video.js
```diff
@@ -84,16 +84,19 @@ export default class HTML5Video extends Playback {
...
+      preload: preload || 'metadata',
+      controls: playbackConfig.preload || options.useVideoTagDefaultControls,
+      crossorigin: playbackConfig.crossorigin
```
tjenkinson: The property is `crossOrigin` not `crossorigin`

src/playbacks/html5_video/html5_video.js
```diff
@@ -84,16 +84,19 @@ export default class HTML5Video extends Playback {
...
+    })
```
tjenkinson: This works different to be before in that it is setting the properties on the video tag instead of adding attributes to the dom element which I guess is fine, except for poster which looks like it can only be set as an attribute.

README.md
```diff
@@ -125,6 +125,22 @@ In case you're loading a on demand video (mp4), it's possible to define the way
 ##### HLS Buffer Length
 The default behavior for the HLS playback is to keep buffering indefinitely, even on VoD. This replicates the behavior for progressive download, which continues buffering when pausing the video, thus making the video available for playback even on slow networks. To change this behavior, add `maxBufferLength: <value>` to embed parameters, where `value` is in seconds.
 
+##### Playback configuration
+The configuration for the playback, it's still only compatible with `html5_video` playback.
+
+```javascript
+{
+  playbackConfig: {
+    loop: 'loop',
```
tjenkinson: should be a boolean

README.md
```diff
@@ -125,6 +125,22 @@ In case you're loading a on demand video (mp4), it's possible to define the way
...
+    autoplay: 'autoplay',
```
tjenkinson: boolean :P

README.md
```diff
@@ -125,6 +125,22 @@ In case you're loading a on demand video (mp4), it's possible to define the way
...
+    poster: 'http://example.com/example.png',
+    preload: 'metadata',
+    controls: 'controls',
```
tjenkinson: boolean :P

towerz:
Looks good to me. I agree with @tjenkinson, the flags should be passed as boolean values, and the case should match our properties.

About the autoplay flag, I think it should be removed, since we already have an autoPlay parameter for the player instance, and the playbacks already receive those in the options object. Having a playback specific autoPlay flag is redundant in my opinion.

leandromoreira:
My only concern about `boolean`, while setting attributes on video tag is that setting `controls: false` does not imply the lack of controls ;/

```html
<video controls=false src="http://www.sample-videos.com/video/mp4/360/big_buck_bunny_360p_2mb.mp4"></video>
```

towerz:
Great job, @leandromoreira! 🍻

towerz:
There's no problem with the autoplay flag, because it uses the value from options, so it's not redundant anymore. 

 About the `crossOrigin`, you're right, I missed that. Since it's just a typo, I'll fix it right away, since I just merged the PR. There's another redundant parameter, `poster`, that needs to be removed.

towerz:
@tjenkinson about the `.poster` property, I believe it does exist and is documented for HTMLVideoElement, at least here: https://developer.mozilla.org/en-US/docs/Web/API/HTMLVideoElement

towerz:
I just pushed a fix to the `crossOrigin` typo and the redundant `poster`/`loop` options. Those options should still be global, they're not specific for HTML5 video playback.

PR #1008, opened 2016-05-29
Remove old baseUrl references and provide way to set baseUrl at runtime

tjenkinson:
It looked like where baseUrl was referenced before it never had any effect, as this changed when we switched to webpack.

The proper way of setting the assets base url at runtime is in the webpack documentation:
https://webpack.github.io/docs/configuration.html#output-publicpath.

I have added an extra entry point which looks to see if the user wants a custom base url, and sets it if they do before the main app loads.
This uses a global because I don't think there's another way.

### DIFF
--- a/README.md
+++ b/README.md
@@ -244,7 +244,7 @@ The project is on npm at https://www.npmjs.com/package/clappr
 By default Clappr will assume that assets are located at the same location as the script.
 If this is not the case you need to set your base url immediately BEFORE requiring Clappr, or importing the script.
 
-```javascript
+```html
   <script type="text/javascript">
     window.CLAPPR_ASSETS_BASE_URL = "http://mycdn.example.com/assets/clappr";
   </script>
--- a/README.md
+++ b/README.md
@@ -241,15 +241,22 @@ The project is on npm at https://www.npmjs.com/package/clappr
 
 `npm install clappr --save-dev`
 
-You should specify the base url for where the assets are located using the `baseUrl` option:
+By default Clappr will assume that assets are located at the same location as the script.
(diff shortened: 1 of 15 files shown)


towerz:
I don't think this change would add much value, since it replaces a runtime parameter used when embedding the player with a global variable that needs to be set before evaluating the Clappr script.

The `baseUrl` parameter exists so that products embedding Clappr could have whatever URL layout they want, allowing separating the javascript files from assets, or whatever the user would want, without the need to build Clappr itself. The assumption that having a `baseUrl` parameter is wrong due to the way webpack works with runtime assets is not correct, since Clappr was designed to not rely on whatever technology is used to build it (even ES6 - we have ES5 helpers for extending classes). It should be simple to extend and transparent to use, and it should not depend on products that include Clappr relying on using webpack as well, or even adding global variables for setting this, which seems really clumsy.

There's also a plan of splitting Clappr into smaller packages, a core and independent plugin bundles, as well as a full player package that includes most plugins that are used today. I'm not sure this change will fit into this future release, so I wouldn't recommend merging it.

leandromoreira:
I also think that the benefits acquired from it is not so great ;/ (given that we'll introduce some bootstrap and rely on global window object ...)

towerz:
@tjenkinson about asynchronous modules, I think we can provide guidance on how to support this the webpack way. 

As I mentioned before, we are not commited to webpack itself, so I don't think Clappr should be altered only to support a webpack feature. If an user wants to use Clappr this way, I think it would be reasonable to just override the default Clappr webpack configuration and provide a valid `publicPath`.

In our custom globo.com player, we override the Clappr build settings to provide a fixed and valid publicPath, so I don't think it would be an obstacle for using asynchronous modules.

Issue #1003, opened 2016-05-26
Unable to run istanbul-mocha tests when referencing clappr


iongion:
Using clappr *0.2.52*

The command I use to start the tests:
```
node_modules/.bin/babel-node ./node_modules/babel-istanbul/lib/cli cover node_modules/mocha/bin/_mocha -- --compilers js:babel-core/register --bail --sort --reporter spec --require test/bootstrap.js --recursive test/unit/**/**/*.spec.js
```

But they suddenly stop running with:
```
W:\Workspace\app\node_modules\clappr\src\main.js:5
import Player from 'components/player'
^^^^^^

SyntaxError: Unexpected token import
    at exports.runInThisContext (vm.js:53:16)
    at Module._compile (module.js:387:25)
    at Module._extensions..js (module.js:422:10)
    at Object.require.extensions.(anonymous function) [as .js] (W:\Workspace\app\node_modules\babel-register\lib\node.js:166:7)
```

Which looks as if babel is not being used.
My *.babelrc* has nothing special:

```
{
  "presets": [
    "es2015",
    "react"
  ],
  "env": {
    "development": {
      "presets": ["react-hmre"]
    }
  },
  "plugins": [
    [
      "babel-root-import",
      {
        "rootPathPrefix": "~",
        "rootPathSuffix": "sources/modules"
      }
    ],
    [
      "babel-plugin-transform-builtin-extend",
      {
        "globals": [
          "Error"
        ]
      }
    ]
  ]
}
```

Do you have any ideas why this happens ?

iongion:
In fact, just running the tests with babel-node 

```
node_modules/.bin/babel-node test/unit/modules/app/EntryPoint.spec.js

import Player from 'components/player'
^^^^^^

SyntaxError: Unexpected token import
    at exports.runInThisContext (vm.js:53:16)
    at Module._compile (module.js:387:25)
```

Showing that the loader is unable to import clappr as ES6 module, what can be done ?

iongion:
Apparently, it happens due to the fact the the babel transpiler does not transpile anything in node_modules so clappr's main.js is not transpiled, that is why the import error.

In the end, the mocha command is altered a bit:
```
mocha -- --bail --sort --require test/bootstrap.js --reporter spec --recursive test/unit/**/**/**.spec.js
```

So that it bootstraps using a file where babel compiler is registered and its settings configured.
```
require('babel-core/register')({
  ignore: /node_modules\/(?!clappr)/,
});
[...]

PR #1011, opened 2016-05-31
Allow mocking of 'window'

tjenkinson:
refs #1003 

Now that I've done this I'm wondering if it might actually make more sense not to have the "base\mocks" module and expect the developer to mock the global `window` variable when they run their tests instead?@clappr/core ?

This PR allows

```javascript
var Clappr = require("clappr");
Clappr.mocks.window = myFakeWindow;

var player = new Clappr.Player() //etc
```

### DIFF
--- a/src/base/utils.js
+++ b/src/base/utils.js
@@ -4,6 +4,7 @@
 /*jshint -W079 */
 
 import Browser from 'components/browser'
+import mocks from 'base/mocks'
 import $ from 'clappr-zepto'
 
 function assign(obj, source) {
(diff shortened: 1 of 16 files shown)


src/base/utils.js
```diff
@@ -195,23 +196,31 @@ export function isNumber(value) {
 }
 
 export function currentScriptUrl() {
```
tjenkinson: Need to figure out how to do this before this. It needs to be called in the same tick as the script being loaded, but can't be called before the developer has a chance to set their `window` :/

PR #1017, opened 2016-06-05
Cap webworkify-webpack to 1.0.6

### LINKED ISSUE #1015: hls play error Uncaught TypeError: Cannot read property 'call' of undefined
when i use the original clapper.min.js it's ok to play hls,
but when i use npm run release to generate my own clapper.min.js then it's throw error as the title

it seems that the hls.js lib  can't execute right so hls can't play .
i want to know how you guys to package the clappr.min.js ,why i use the original file can work fine ,
but use my own generate file can't work.

@leandromoreira ,can you help me ,thanks a lot!

tjenkinson:
Until can figure out root issue.

Fixes #1015

### DIFF
--- a/package.json
+++ b/package.json
@@ -78,7 +78,7 @@
     "url-loader": "^0.5.6",
     "webpack": "^1.12.1",
     "webpack-dev-server": "^1.14.0",
-    "webworkify-webpack": "^1.0.4",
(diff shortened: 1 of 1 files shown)


leandromoreira:
LGREAATOMe

PR #1019, opened 2016-06-07
Delete rx.lite.compat.js

thiagopnts:
I guess this is not used anywhere


leandromoreira:
I really would like to use it 💃

PR #1023, opened 2016-06-08
adds new event play_intent

leandromoreira:
While working on a new [plugin](https://github.com/leandromoreira/clappr-stats) I tried to estimate the `startup` time but I couldn't, because the `PLAY` event happens after the first `BUFFERING` :P also @towerz is working in some things that will require this event.


towerz:
:+1:

PR #1027, opened 2016-06-12
Organize the README

leandromoreira:
***1 Upvote*** The idea is to split and link the content (and I also added the troubleshooting guide), we can bring only the most important content to the home page.


INSTALLING.md
```diff
@@ -0,0 +1,51 @@
...
+  var player = new Clappr.Player({
+  	source: "http://your.video/here.mp4",
+	baseUrl: "http://example.com/assets/clappr"
+  });
+```
+In the above case clappr will expect all of the [assets (in the dist folder)](https://github.com/clappr/clappr/tree/master/dist) to be accessible at "http://example.com/assets/clappr".
+You need to arrange for the assets to be located at `baseUrl` during your build process.
+
+#### Installing for [webpack](https://webpack.github.io/)
+By default webpack will look at the `browser` field in `package.json` and use the built version of the project. If this is all you want there is nothing else for you to do.
```
tjenkinson: This should now be

By default webpack will look at the `main` field in `package.json`

INSTALLING.md
```diff
@@ -0,0 +1,51 @@
...
+```javascript
+resolve: {
+    alias: { Clappr: 'clappr/src/main.js' },
+    root: [path.resolve(__dirname, 'node_modules/clappr/src')],
+    extensions: ['', '.js'],
+}
+```
+
+#### Installing for [browserify](http://browserify.org/)
+Browserify will look at the `browser` field in `package.json` and use the built verison of the project.
```
tjenkinson: and `main` here as well.

flavioribeiro:
<3

PR #1037, opened 2016-06-21
adds a plugin showcase

leandromoreira:
***1 Upvote*** **DO NOT ACCEPT THIS PR YET** we need to

* create an art for each plugin (280x160)
* order/classify them 

https://github.com/clappr/clappr/blob/b1a70cfe3ec4162221734e8ee302f1ba408a2634/README.md#external-plugins What are your thoughts?


thiagopnts:
nice 💯

PR #1047, opened 2016-06-25
avoids call destroy twice

leandromoreira:
I noticed that calling `destroy` on core calls `destroy` twice on playback.

https://github.com/clappr/dash-shaka-playback/issues/14

Do you see any problem in this fix? I mean, only if the plugins order would matter, if we destroy a X plugin before the Playback.


towerz:
I don't think there's any problem with this approach, the playback is always added to the container plugin list. Since it's handled together with all other container plugins, there should be no problem destroying it with the other plugins.

PR #1064, opened 2016-07-05
Add type="button" to prevent form submission from embedded player

bikegriffith:
***2 Upvotes*** If the player is embedded inside a `<form>` element (e.g. on an ASP.NET site where the whole page is a giant form), trying to press one of the media control buttons could fire the form submission unexpectedly, which results in strange behavior on full-screen, back to live, etc.

By changing the button type to "button" (instead of default of "submit"), tapping the button does not submit the form any more.


leandromoreira:
👍

PR #1069, opened 2016-07-06
Extrapolate start time on hls playback

### LINKED ISSUE #1022: Cursor jitter in DVR playback
B**Browser**: Chrome 51.0.2704.84

**OS**: Mac OS X 10.11.5

**Clappr Version**: 0.2.54

**Steps to reproduce**:

* launch this [test page ](http://streamroot.test.s3.amazonaws.com/3rdparty/clappr/clappr-level-selector.html)  It is latest clappr with latest level-selector plugin loading an MBR HLS stream with a 5 minute DVR playlist.

* move the playback cursor back to enter the DVR buffer (e.g. 1 minute back) and observe the cursor progress in playback mode

* The cursor should remain roughly at the same position since the playlist is a fixed length 5 minute buffer (DVR mode) instead of that the cursor jitters back and forth during playback. 

Did you try to reproduce this issue at http://cdn.clappr.io/

yes.

tjenkinson:
Increases the start time in sync with the user's clock, and tweaks when necessary to remain in sync with remote playlist.

This means cursor remains still during a live playlist with a sliding window.

For VOD or live playlists of type EVENT the start time will always be the start time of the first segment, which should be 0 (when https://github.com/dailymotion/hls.js/pull/547 is merged).

Closes #1022 

Waiting on https://github.com/dailymotion/hls.js/pull/547


src/playbacks/hls/hls.js
```diff
@@ -28,12 +29,73 @@ export default class HLS extends HTML5VideoPlayback {
...
     super(...args)
     // backwards compatibility (TODO: remove on 0.3.0)
     this.options.playback || (this.options.playback = this.options.hlsjsConfig)
-    this._minDvrSize = (this.options.hlsMinimumDvrSize === undefined) ? 60 : this.options.hlsMinimumDvrSize
+    this._minDvrSize = typeof(this.options.hlsMinimumDvrSize) === "undefined" ? 60 : this.options.hlsMinimumDvrSize
+    // The size of the start time extrapolation window measured as a multiple of segments.
+    // Should be 2 or higher, or 0 to disable. Should only need to be increased above 2 if more than one segment is
+    // removed from the start of the playlist at a time. E.g if the playlist is cached for 10 seconds and new chunks are
+    // added/removed every 5.
+    this._extrapolatedWindowNumSegments = typeof(this.options.playback.extrapolatedWindowNumSegments) === "undefined" ? 2 :  this.options.playback.extrapolatedWindowNumSegments
```
leandromoreira: @towerz are we going to use `options.playback` or `options.hls` ?

src/playbacks/hls/hls.js
```diff
@@ -28,12 +29,73 @@ export default class HLS extends HTML5VideoPlayback {
...
+  }
+
+  get _startTime() {
+    if (this._playbackType === Playback.LIVE && this._playlistType !== "EVENT") {
+      return this._extrapolatedStartTime
+    }
+    return this._playableRegionStartTime
+  }
+
+  get _now() {
```
leandromoreira: @clappr/core should we move this to core? I think it might be useful for other stuffs.. although it could also be a premature optimization

towerz: @leandromoreira perhaps we should move this to utils. It could be useful for other components.

src/playbacks/hls/hls.js
towerz: @leandromoreira as we discussed before, we should stick to `options.playback` for playback-specific options, and `options.pluginName` for core/container plugin-specific options.

I don't think we really need to have options separated by playback name, because this way we can use the same option name/value for any playback used for a given source. For example, we could use the same option name for hls.js and FlasHLS playbacks.

src/playbacks/hls/hls.js
tjenkinson: Yes `get _now()` could move but sure about `_duration()` and `_startTime()` because they both use properties that are specific to this playback.

leandromoreira: I think only `get now`!

tjenkinson: :+1:

leandromoreira:
@tjenkinson there are some conflicts to be solved

thiagopnts:
looks like this increases 16kb in the minified player size, which is already around ~410kb, which is almost double the size of the player a couple months ago. I think we need to pay more attention to this

towerz:
@tjenkinson will you try to remove the lodash dependencies? Or are we ok to merge it as is?

@thiagopnts I'm aware that the bundle size has increased, but I'm still working on restructuring plugins and the media control to release a new version of Clappr without bundling all plugins with it. This way we can have more control of the player size, as well as allowing loading plugins separately.

Issue #1076, opened 2016-07-12
Incorrect buffer events with autoplay


shaharmor:
When `autoplay` is true, there buffer start event is not called when it should.
This also causes the spinning wheel to not appear on initial load.

The line that is responsible for "starting" the buffer event is this: https://github.com/clappr/clappr/blob/master/src/playbacks/html5_video/html5_video.js#L180
But if autoplay is true, instead of calling the `.play` function like normal playback, this line: https://github.com/clappr/clappr/blob/master/src/playbacks/hls/hls.js#L52 is calling `_setupHls()` directly, skipping the logic inside `.play()` basically, which screws the buffer start event firing

I suggest to just call `this.play()` instead of calling `_setupHls()` directly

**Clappr Version**: 0.2.57

tjenkinson:
@shaharmor I think this might be a sightly different issue because

- `play()` is never called when `autoPlay` is `true`. Is the same issue present for a standard mp4 with autoPlay?
- `_setupHls()` also doesn't trigger playback to start. It just configures hlsjs. The reason playback starts after `_setupHls()` has completed is because the `autoplay` attribute is on the video tag. We could always call `setupHls()` immediately but we chose not to to prevent buffering occurring if the user hasn't requested play yet.
- As soon as the 'playing' event is emitted `_handleBufferingEvents()` should be called.
https://github.com/clappr/clappr/blob/master/src/playbacks/html5_video/html5_video.js#L279

shaharmor:
@tjenkinson I didn't check mp4.

Regarding this line you sent: https://github.com/clappr/clappr/blob/master/src/playbacks/html5_video/html5_video.js#L279
it will be called the `playing` is emitted, but the issue i'm talking about it the **initial** buffer-start event.

In normal conditions i expect `CONTAINER_STATE_BUFFERING` to be called as soon as the initial loading starts (when the first request to the manifest is sent basically), but if `autoPlay` is true, it will only be called when playback **starts**, which is basically incorrect.

shaharmor:
Btw, this also means that if autoPlay is true, no spinning/loading gif will be shown when loading starts, which is also a problem no?

shaharmor:
I still think that calling `.play` is the right way, (Maybe only for hls playback).
Btw, in flash, you do call `this.play()`.
Why not call `.play` directly instead of using the `autoPlay` property of the video tag?

tjenkinson:
[...]

PR #1077, opened 2016-07-12
Html5Video: explicit call play() when autoPlay option

tjenkinson:
video.paused remains true initially even if autoplay attribute set, but flips to false immediately after a play() call.
Refs #1076


towerz:
Took some time to test whether this would affect anything else. Seems ok to merge this.

Perhaps we should try to increase playback state test coverage so we don't have these issues again.

towerz:
Thanks, @tjenkinson!

PR #1078, opened 2016-07-13
Allow disable audio checking

DerivcoHK:
Hi,
I'm using clappr to play hls video.
However, the isAudio detection decide it will play in audio tag.

I add a configuration that allow developer to skip checking is audio or not.

Thx


towerz:
Hi, @DerivcoHK. We could not reproduce this behavior you reported. I even added a test to check whether the `isAudioOnly` is returning a false positive (84318ce2b641b96116129be7d1a6e2273ca9e858).

If it's a bug, it's better if we solve the problem with the `isAudioOnly` check, instead of adding a flag to disable it. Could you please open an issue reporting the problem and how we can reproduce this behavior? 

I'll close this PR, since it's not considered an ideal solution to the possible bug you found.

Issue #1082, opened 2016-07-14
Fullscreem inside iframe without allowfullscreen


enbplas11:
Good people, I'm new to this good player but I have a drawback.

I can do fullscreen on a iframe without using the tag allowfullscreen inside the iframe.

What I need is this:

`<Iframe src = "http://cdn.clappr.io"> </ Iframe>`

I'm using the version clappr-0.2.58

I need your help

leandromoreira:
Hi @enbplas11 , what issue are you facing? Do you want clappr to disable fullscreen button?

tjenkinson:
That's a good idea and we could actually add detection for this.

https://github.com/LA1TV/Website/blob/master/app/assets/src/app/pages/embed/player-page.js#L21-L29

enbplas11:
Hi @leandromoreira , What happens is that I have a page with clappr where users perform an iframe my page and it does not work the fullscreem.

by the number of users who use my site can not add  tag and so do not lose traffic, I currently use another player that if I do fullscreem without adding the tag (jwplayer 6) allowfullscreem within the iframe.
I was reviewing a issue: [https://github.com/clappr/clappr/commit/0476dc6984b9700247cb1d03a9624526df1de81e](url) but not for the latest version

What solution could be added to this version of clappr without using the allowfullscreem inside the iframe?

tjenkinson:
@enbplas11 please can you provide a link to a demo page with your problem? AFAIK content cannot go full screen from inside an iframe unless the user has given explicit permission by adding the tag to the iframe. It's a browser restriction. Would be interested to see how jwplayer gets around this.

leandromoreira:
>FAIK content cannot go full screen from inside an iframe unless the user has given explicit permission by adding the tag to the iframe.
Same

enbplas11:
Hi @tjenkinson , This page [http://cinestrenostv.tv/canales/envivo/antena3.html](url) makes the call to this page [http://www.ponlatv.com/channel.php?file=103950&width=650&height=400&autostart=true](url) currently uses the jwplayer 6.

This is the link that implements: [http://www.ponlatv.com/channel_nuevo.php?file=105749&width=680&height=390&autostart=true](url)
but if another user makes an iframe to that url the fullscreem not work, as shown in the following link: [http://www.ponlatv.com/iframe_release2.php](url)

tjenkinson:
:+1: makes sense

leandromoreira:
[...]

Issue #780, opened 2016-01-19
Disable fullScreen and option to change it to fullWindow


nbashev:
I'm not able to disable fullScreen button. The only way to disable it is via delete from template. Is there an option to deactivate fullScreen and may be an option to alter it with fullWindow. Most Andorid devices will not accept custom layout over the player in full screen. With the help of an app (browser without url bar) we can emulate fullScreen and keep the UI part of Clappr in tact.

PR #1085, opened 2016-07-15
hide fullscreen button if full screen not possible

tjenkinson:
E.g If player is inside an iframe without the `allowFullScreen` attribute.

Refs https://github.com/clappr/clappr/issues/1082#issuecomment-232637828

### DIFF
--- a/src/base/utils.js
+++ b/src/base/utils.js
@@ -86,7 +86,7 @@ export var Fullscreen = {
       document.msExitFullscreen()
     }
   },
-  canGoFullscreen: function() {
+  fullscreenEnabled: function() {
     return !(
       document.fullscreenEnabled === false ||
       document.webkitFullscreenEnabled === false ||
--- a/src/base/utils.js
+++ b/src/base/utils.js
@@ -85,6 +85,14 @@ export var Fullscreen = {
     } else if(document.msExitFullscreen) {
       document.msExitFullscreen()
     }
+  },
+  canGoFullscreen: function() {
+    return !(
+      document.fullscreenEnabled === false ||
+      document.webkitFullscreenEnabled === false ||
+      document.mozFullScreenEnabled === false ||
+      document.msFullscreenEnabled === false
+    )
   }
 }
 
@@ -212,6 +220,14 @@ export function getBrowserLanguage() {
   return window.navigator && window.navigator.language
 }
 
+// remove the item from the array if it exists in the array
+export function removeArrayItem(arr, item) {
+  var i = arr.indexOf(item)
+  if (i >= 0) {
+    arr.splice(i, 1)
+  }
+}
+
 export default {
   Config,
   Fullscreen,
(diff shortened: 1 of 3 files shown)


src/base/utils.js
```diff
@@ -85,6 +85,14 @@ export var Fullscreen = {
...
       document.msExitFullscreen()
     }
+  },
+  canGoFullscreen: function() {
+    return !(
+      document.fullscreenEnabled === false ||
+      document.webkitFullscreenEnabled === false ||
+      document.mozFullScreenEnabled === false ||
+      document.msFullscreenEnabled === false
+    )
```
towerz: Why don't you instead of this do:
```javascript
return !!(document.fullscreenEnabled ||
  document.webkitFullscreenEnabled ||
  document.mozFullScreenEnabled ||
  document.msFullscreenEnabled
)
```
I think it reads better, and avoids comparing to `false`.

src/base/utils.js
```diff
@@ -85,6 +85,14 @@ export var Fullscreen = {
     } else if(document.msExitFullscreen) {
       document.msExitFullscreen()
     }
+  },
+  canGoFullscreen: function() {
```
towerz: I prefer using a standards based name, like `fullscreenEnabled` or `isFullscreenEnabled`. Why don't we go this way?

leandromoreira:
LGTM once again thanks mister @tjenkinson 🍻

src/base/utils.js
tjenkinson: It is to make sure if they are all undefined it defaults to assuming full screen is available. Can change it though if we think one those properties will always exist?

src/base/utils.js
tjenkinson: Yes that makes more sense I'll update it

src/base/utils.js
towerz: If none of those properties are available, it means that the fullscreen API is unavailable. It would be incorrect to assume that the player can enter fullscreen in this case.

towerz: Actually, we had an idea some time ago to maximize the player in the window viewport if the fullscreen API is not available. This way we can support maximizing the player size on some mobile browsers that do not support fullscreen. Perhaps we could create an issue to discuss this idea. What do you think, @tjenkinson?

In the meantime, we can avoid having false positives assuming `undefined` values as unavailable.

tjenkinson: Sounds like a good feature to me. I guess
```
position: fixed;
top: 0;
left: 0;
bottom: 0;
right: 0;
```
?

towerz: It's not as simple as that, because it has to fill all available viewport space and display on top of everything in the page. If some other element has an increased z-index, we should render on top of it as well.

towerz: I was creating an issue so we can discuss that, and found #780. Since this PR solves the issue partially, I'll leave this referenced and create another one so we can approach the full viewport request.

towerz:
Merged, thanks @tjenkinson!

PR #1086, opened 2016-07-18
Removing duplicate event listener from google_analytics plugin

stanogurnik:
Just simple change: Listener `Events.CONTAINER_ENDED` is registered twice in Google Analytics plugin.


leandromoreira:
Thanks @stanogurnik

PR #1091, opened 2016-07-22
Match by lowercase extensions in "_canPlay"

sjlu:
In case the URL contains a uppercase character, it'll help match the mediaType list (e.g. "MP4" vs "mp4)

### DIFF
--- a/src/playbacks/html5_video/html5_video.js
+++ b/src/playbacks/html5_video/html5_video.js
@@ -408,7 +408,7 @@ export default class HTML5Video extends Playback {
 
 HTML5Video._canPlay = function(type, mimeTypesByExtension, resourceUrl, mimeType) {
   var extension = (resourceUrl.split('?')[0].match(/.*\.(.*)$/) || [])[1]
-  var mimeTypes = mimeType || mimeTypesByExtension[extension] || []
+  var mimeTypes = mimeType || (extension && mimeTypesByExtension[extension.toLowerCase()]) || []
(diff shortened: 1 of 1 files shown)


src/playbacks/html5_video/html5_video.js
```diff
@@ -408,6 +408,8 @@ export default class HTML5Video extends Playback {
 
 HTML5Video._canPlay = function(type, mimeTypesByExtension, resourceUrl, mimeType) {
   var extension = (resourceUrl.split('?')[0].match(/.*\.(.*)$/) || [])[1]
+  extension = (extension || "").toLowerCase()
+  
   var mimeTypes = mimeType || mimeTypesByExtension[extension] || []
```
tjenkinson: Hi @sjlu

Well spotted thanks for the PR.

I think changing this line to
```javascript
var mimeTypes = mimeType || (extension && mimeTypesByExtension[extension.toLowerCase()]) || []
```
would be better, because we shouldn't be trying to access `mimeTypesByExtension[undefined]`

sjlu: Agreed, will change.

sjlu:
@tjenkinson change made

sjlu:
@tjenkinson #1094 👍

towerz:
Thanks @sjlu! :beers:

sjlu:
Glad to be of help @towerz

Issue #1092, opened 2016-07-22
Audio only playback, no video


sjlu:
**Browser**: Chrome 51.0.2704.103 (64-bit)
**OS**: OS X 10.11.5
**Version**: 0.2.59

**Steps to reproduce**:

* set opts.mimeType to "video/mp4"
* set URL with an invalid or no proper extension
* player plays as audio only due to "_canPlay" reporting true for "isAudioOnly"
* replicable on CDN version

e.g. http://gameclipscontent-d2018.xboxlive.com/00090000095ab1d1-802dee60-54e2-408c-9b94-c97c92f780a8/GameClip-Original.MP4?sv=2014-02-14&sr=c&sig=FCt61jzzj0DVIZm%2FSqTb1HShP566M2ojwDhduGpwdWM%3D&st=2016-07-22T18%3A55%3A49Z&se=2016-07-22T20%3A00%3A49Z&sp=r&__gda__=1469217649_2ba9d477776132e714dc370686c1e352

**Debugging**:
* [src/playbacks/html5_video/html5_video.js#L42](https://github.com/clappr/clappr/blob/master/src/playbacks/html5_video/html5_video.js#L42) `AudioOnly` returns `true` due to a set of parameters sent to `HTML5Video._canPlay` 
* [src/playbacks/html5_video/html5_video.js#L409](https://github.com/clappr/clappr/blob/master/src/playbacks/html5_video/html5_video.js#L409) `type = "audio"` `mimeTypes = ["video/mp4"]`
* [src/playbacks/html5_video/html5_video.js#L415](https://github.com/clappr/clappr/blob/master/src/playbacks/html5_video/html5_video.js#L409)
`media.canPlayType("video/mp4") => "maybe"`

sjlu:
Possibly related to #1000

tjenkinson:
I think maybe instead of 
```javascript
HTML5Video._canPlay('audio', AUDIO_MIMETYPES, resourceUrl, mimeType)
```
it should be
```
!HTML5Video._canPlay('video', VIDEO_MIMETYPES, resourceUrl, mimeType)
```
in https://github.com/clappr/clappr/blob/master/src/playbacks/html5_video/html5_video.js#L42

sjlu:
As I was walking away, I was thinking exactly the same thing. Could also be a combination of the two - will make that change too.

towerz:
@tjenkinson actually, @sjlu solution is right. Using only 
```javascript
[...]

PR #1093, opened 2016-07-22
ensuring that isAudioOnly does not pass true for video types

sjlu:
Making sure that 'video' mimetype does not pass validation in `isAudioOnly()` function.


sjlu:
Related to #1092

towerz:
Thanks @sjlu! :beers:

PR #1094, opened 2016-07-22
match lowercase based extensions

sjlu:
Several changes to help match against extensions that happen to be uppercased.


towerz:
Thanks @sjlu! :beers:

PR #1110, opened 2016-08-08
Fix for flexbox container layout issues on older browsers

michaelcunningham19:
Confirmed affected browsers:
Internet Explorer 10
Safari on iOS 8.4 (iPad Air 2)

The container for the player poster did not have the Play button vertically aligned in the center.

Before:
https://www.dropbox.com/s/2hoae0yssccroc5/Screenshot%202016-08-08%2014.14.27.png?dl=0

After:
https://www.dropbox.com/s/ujv85f01xi75hzt/Screenshot%202016-08-08%2014.14.01.png?dl=0

When using flexbox with an older browser that implements an older version of the flexbox specification, prefixes must be added, in some cases, default values are different as well. See link:
https://msdn.microsoft.com/en-us/library/hh673531(v=vs.85).aspx

Prefixed values were generated with Autoprefixer: https://autoprefixer.github.io/

### DIFF
--- a/src/plugins/poster/public/poster.scss
+++ b/src/plugins/poster/public/poster.scss
@@ -1,21 +1,15 @@
 @import "bourbon";
 
 .player-poster[data-poster] {
+  @include display(flex);
+  @include justify-content(center);
+  @include align-items(center);
   position: absolute;
   height: 100%;
   width: 100%;
   z-index: 998;
   top: 0;
   left: 0;
-  display: -webkit-box;
-  display: -ms-flexbox;
-  display: flex;
-  -webkit-box-pack: center;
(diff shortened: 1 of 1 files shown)


towerz:
@tjenkinson not really, we already use bourbon. We should use bourbon mixins, as we are already using on other plugins. This was only an oversight.

@michaelcunningham19 could you please update the PR using proper [bourbon mixins](http://bourbon.io/docs/#flexbox) instead of expanding the prefixes?

michaelcunningham19:
@towerz 

I have updated the PR to use bourbon mixins 👍

towerz:
Thanks, @michaelcunningham19! 🍻

PR #1111, opened 2016-08-08
Improve no-op playback styling

tjenkinson:
Wasn't perfectly centred when using margin. Works when using padding.

Overflow and max-height setting means if the message is longer than the player the user will be able to scroll to see the entire message.


leandromoreira:
LGTM

towerz:
Thanks @tjenkinson! :beers:

PR #1112, opened 2016-08-08
Handle exceptions in early event handler

### LINKED ISSUE #751: Handle exceptions thrown in event handlers
Given event listeners may come from third party plugins I think it would make sense for the 'Events' class to catch any exceptions thrown from a listener, log them, and move on to the next listener instead of letting the exception traverse further up.

Related: https://github.com/clappr/clappr-level-selector-plugin/pull/21

tjenkinson:
Also checks that handlers are called in the order they were registered
Closes #751

### DIFF
--- a/src/base/events.js
+++ b/src/base/events.js
@@ -33,15 +33,24 @@ const eventsApi = function(obj, action, name, rest) {
   return true
 }
 
-const triggerEvents = function(events, args) {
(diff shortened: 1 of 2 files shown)


leandromoreira:
AWESOME : ) LGTM

leandromoreira:
Thanks mister @tjenkinson lol

PR #1133, opened 2016-08-27
Check if fullscreen supported on <video /> if other fullscreen checks fail

### LINKED ISSUE #1127: Fullscreen option is not available on iPad
**Browser**: Safari for iOS 9.3

**OS**: iOS 9.3

**Clappr Version**: 0.2.25

**Steps to reproduce**:
* open cdn.clappr.io on iPad
* play video
* observe that media control does not have the fullscreen button

**Did you try to reproduce this issue at http://cdn.clappr.io/**
Yes

**Other Notes**
It looks like `document.webkitFullscreenEnabled` is not available on Safari for iOS.  However, fullscreen capabilities do exist.

Note that you can double-tap the video to have it go fullscreen still.

The fullscreenEnabled check that works for iOS Safari would have to wait for the media to load, and then look at `videoElement.webkitSupportsFullscreen` (which is not available until the media is loaded)

https://developer.apple.com/library/safari/documentation/AudioVideo/Conceptual/Using_HTML5_Audio_Video/ControllingMediaWithJavaScript/ControllingMediaWithJavaScript.html

tjenkinson:
This is required for iOS
fixes #1127

### DIFF
--- a/src/components/media_control/media_control.js
+++ b/src/components/media_control/media_control.js
@@ -18,6 +18,7 @@ import template from 'base/template'
 import Playback from 'base/playback'
 
 import $ from 'clappr-zepto'
+import merge from 'lodash.merge'
 
 import mediaControlStyle from './public/media-control.scss'
 import mediaControlHTML from './public/media-control.html'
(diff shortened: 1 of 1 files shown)


leandromoreira:
lgtm

bikegriffith:
Per my comment in the issue, this solves the original problem. Tested on an iPad Air and iPad 2 with iOS 9.3

leandromoreira:
LGTM @clappr/core thoughts?

leandromoreira:
Thanks mister @tjenkinson lol

PR #1141, opened 2016-09-06
favicon: Remove event listeners from old container on change

tjenkinson:
previously the event listeners were removed off the new container which has no effect


towerz:
Great job, thanks @tjenkinson!

PR #1142, opened 2016-09-07
read settings from one place and add getSettings()

tjenkinson:
`getSettings()` can then easily be overridden by someone who wants to extend `MediaControl` and change its behaviour.

E.g override `seekEnabled`  to false if showing an ad. In the custom implementation they can also call `settingsUpdate()` manually if they change a setting.

In order to create an ad plugin which would disable the seek bar (and possibly other buttons) the media control class could be extended like so:

``` javascript
class MyMediaControl extends MediaControl {
  constructor(options) {
    super(options)
    this._adModeEnabled = false
  }

  enableAdMode(enable) {
    this._adModeEnabled = enable
    this.settingsUpdate()
  }

  getSettings() {
    let newSettings = $.extend({}, this.container.settings)
    if (!this._adModeEnabled) {
      newSettings.seekEnabled = false
    }
    return newSettings
  }
}
```

And then a plugin could access the media control and call the `enableAdMode()` method.

There might be a better way to achieve this @clappr/core ?
@vix-simplex mentioned in gitter that he'd like to be able to do this


leandromoreira:
Thanks @tjenkinson I like the idea I'd just change `getX` to `get settings` also @towerz is working on the redesign of media control... I'd like his thoughts on this!

leandromoreira:
@tjenkinson good catch.

towerz:
@tjenkinson could you update this PR to solve conflicts?

towerz:
Thanks @tjenkinson! 👍

PR #1156, opened 2016-09-28
Add missing dvrEnabled getter in flashls playback.

kslimani:
Without this getter, it always return false because of `!!` operator in [isDvrEnabled()](https://github.com/clappr/clappr/blob/master/src/components/container/container.js#L168) container method.

Tested on Internet Explorer 11 with Windows 7. _(which use flash hls instead of hls.js)_

### DIFF
--- a/src/playbacks/flashls/flashls.js
+++ b/src/playbacks/flashls/flashls.js
@@ -675,6 +675,10 @@ export default class FlasHLS extends BaseFlashPlayback {
     }
   }
 
+  get dvrEnabled() {
+    return this._dvrEnabled
+  }
+
   _createCallbacks() {
     if (!window.Clappr) {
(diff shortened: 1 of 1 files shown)


src/playbacks/flashls/flashls.js
```diff
@@ -675,6 +675,10 @@ export default class FlasHLS extends BaseFlashPlayback {
     }
   }
 
+  get dvrEnabled() {
+    return this._dvrEnabled
```
tjenkinson: I think `!!this._dvrEnabled` would be better to ensure it's always a boolean, as this starts as `undefined`.

kslimani: i added !! operator. (feel free to squash to single commit)

PR #1159, opened 2016-09-28
WIP - Fix stop()  for 'html5_video' playback - Safari HLS DVR live stream

### LINKED ISSUE #1158: stop/pause button is bugged with live dvr hls stream on iOS with Safari
If you play hls stream with dvr capabilities on iOS with safari and stop video, if you play again, it start from the begining and there is no way to go back to live (unless reload the source).

I findout this is because Safari use "native" HTMLMediaElement  which is 'html5_video' playback and dvr capabilities detection is not implemented yet. (if an hls source is loaded of course).

Tested with Safari version 10.0 and mac OS X version 10.11.6 (and also safari on iPad)
And Clappr master branch & latest release.

kslimani:
This should fixes #1158.

This is an __experimental fix__, i need feedbacks to ensure it breaks nothing.

### DIFF
--- a/src/playbacks/html5_video/html5_video.js
+++ b/src/playbacks/html5_video/html5_video.js
@@ -192,7 +192,7 @@ export default class HTML5Video extends Playback {
   stop() {
     this.pause()
     this._stopped = true
-    this.el.currentTime = 0
+    this.el.currentTime = this._getStopCurrentTime()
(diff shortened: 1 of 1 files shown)


kslimani:
I close this PR as it it supersede by #1160.

PR #1160, opened 2016-09-28
Html5Video: remove src on stop() and re-add on play()

tjenkinson:
Removed adding `<source>` as I don't think this was necessary. I think the browser would look at the source list to determine what url to assign to the video tag itself, which we are always doing in `setupUrl()`. The video tag itself doesn't have a 'type' attribute which is why I think this.

This might be a solution to #1158 

**Edit:** After reading around a bit more it appears browsers get the type from the `content-type` header returned with the file. `type` exists on the `<source>`'s list so that the browser doesn't have to request each file to get the mime type header (which it would do if the `<source>`'s had the `type` missing).

### DIFF
--- a/src/playbacks/html5_video/html5_video.js
+++ b/src/playbacks/html5_video/html5_video.js
@@ -5,12 +5,10 @@
 import {seekStringToSeconds} from 'base/utils'
 
 import Playback from 'base/playback'
-import template from 'base/template'
 import Styler from 'base/styler'
 import Browser from 'components/browser'
 import Events from 'base/events'
 import tagStyle from './public/style.scss'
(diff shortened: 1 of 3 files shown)


kslimani:
I will test & confirm if it fix #1158 tomorrow at office. (But unless i missed something, it should !)

kslimani:
@tjenkinson @leandromoreira : i tested with

* Safari 10 with Mac OS X El Capitan 10.11.6
* Safari 9 with iPad and iOS 9.2

It fixes #1158 (live HLS source with DVR). However i did not check the `content-type` issue described by @tjenkinson.

PR #1169, opened 2016-10-04
Disable media control and do not show poster when Playback.NO_OP

tjenkinson:
Currently the poster and media control are still shown but are below the no_op div.

PR #1170, opened 2016-10-05
Fix no_op playback autoplay when autoPlay option is set to false

### LINKED ISSUE #1157: no_op playback doesn't respect autoplay option
**Browser**: Chromium 53.0.2785.116

**OS**: Linux 4.7

**Clappr Version**: 0.2.63

**Steps to reproduce**:
- Set autoPlay to false
- Make player use no_op playback
- Load page with the player - player autostarts showing no_op playback with "Your browser is not supported" message

Did you try to reproduce this issue at http://cdn.clappr.io - Yes

I don't know if this supposed to be a feature, but I think it would be more appropriate to respect specifically specified autoplay option and don't autostart the player.

NicholasAsimov:
Implements play() method for no_op playback so it behaves like other playbacks. 
Fixes #1157

### DIFF
--- a/src/playbacks/no_op/no_op.js
+++ b/src/playbacks/no_op/no_op.js
@@ -25,7 +25,10 @@ export default class NoOp extends Playback {
     this.$el.html(this.template({message: playbackNotSupported}))
     this.$el.append(style)
     this.trigger(Events.PLAYBACK_READY, this.name)
-    this._options.autoPlay && this.play()
+    const showForNoOp = !!(this.options.poster && this.options.poster.showForNoOp);
+    if (this.options.autoPlay || !showForNoOp) {
+      this.play()
+    }
     return this
   }
 
--- a/src/playbacks/no_op/no_op.js
+++ b/src/playbacks/no_op/no_op.js
@@ -16,6 +16,7 @@ export default class NoOp extends Playback {
   constructor(...args) {
     super(...args)
     this._noiseFrameNum = -1
+    this.started = false
   }
 
   render() {
@@ -23,9 +24,17 @@ export default class NoOp extends Playback {
     const playbackNotSupported = this.options.playbackNotSupportedMessage || this.i18n.t('playback_not_supported')
     this.$el.html(this.template({message: playbackNotSupported}))
     this.$el.append(style)
-    this._animate()
     this.trigger(Events.PLAYBACK_READY, this.name)
+    this._options.autoPlay && this.play()
     return this
+  }
+
+  play() {
+    if (!this.started) {
+      this.started = true
+      this.trigger(Events.PLAYBACK_PLAY)
+      this._animate()
+    }
   }
 
   _noise() {
--- a/src/playbacks/no_op/no_op.js
+++ b/src/playbacks/no_op/no_op.js
@@ -16,7 +16,7 @@ export default class NoOp extends Playback {
   constructor(...args) {
     super(...args)
     this._noiseFrameNum = -1
-    this.started = false
+    this._started = false
   }
 
   render() {
@@ -30,8 +30,8 @@ export default class NoOp extends Playback {
   }
 
   play() {
-    if (!this.started) {
-      this.started = true
+    if (!this._started) {
+      this._started = true
       this.trigger(Events.PLAYBACK_PLAY)
       this._animate()
     }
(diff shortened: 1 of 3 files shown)


src/plugins/poster/poster.js
```diff
@@ -98,7 +98,9 @@ export default class PosterPlugin extends UIContainerPlugin {
       this.container.disableMediaControl()
       this.$el.show()
     } else {
-      this.container.enableMediaControl()
+      if (this.container.playback.name !== 'no_op') {
+        this.container.enableMediaControl()
+      }
```
tjenkinson: If #1169 is merged then these 3 lines won't be needed as a special case, as even if the media control is enabled it won't appear.
I think it's better if the poster plugin isn't responsible for hiding the media control because some users might not use the poster plugin and the media control should still not be present in that case for the playback type `NO_OP`.

NicholasAsimov: Yep, completely agree. Will do.

kslimani:
@NicholasAsimov : yes, please add an option to keep backward compatibility.

NicholasAsimov:
How should we organize poster options? Right now we have a 'poster' option which changes the image. Maybe we could have a posterOptions object with image and showForNoOp options. I'm not too happy with those names though.

kslimani:
maybe keep `poster` option, and if it is a string, act as poster image path, and if it is an object then check for some expected props  ? (just a thought)

NicholasAsimov:
That could make it backwards-compatible.

leandromoreira:
> That could make it backwards-compatible.

And it's aligned with what we're planning to future.

NicholasAsimov:
I will wait until #1169 is merged before thinking about this more. Currently I feel like it's over complicates the issue.

src/playbacks/no_op/no_op.js
```diff
@@ -25,13 +25,16 @@ export default class NoOp extends Playback {
     this.$el.html(this.template({message: playbackNotSupported}))
     this.$el.append(style)
     this.trigger(Events.PLAYBACK_READY, this.name)
-    this._options.autoPlay && this.play()
+    const showForNoOp = !!(this.options.poster && this.options.poster.showForNoOp);
+    if (this.options.autoPlay || !showForNoOp) {
+      this.play()
+    }
```
tjenkinson: I think it would be better to have the animation always running than have the playback have to look at poster options as the playback and the poster plugin should be completely independent. Also someone might not be using the poster plugin in which case they would also expect it to be animating.

I.e call `this._animate()` in the `render()` method. (I think the `render()` method in a playback is only called once)

src/plugins/poster/poster.js
```diff
@@ -6,14 +6,18 @@ import UIContainerPlugin from 'base/ui_container_plugin'
...
 import posterStyle from './public/poster.scss'
 import posterHTML from './public/poster.html'
 import playIcon from 'icons/01-play.svg'
 
 export default class PosterPlugin extends UIContainerPlugin {
   get name() { return 'poster' }
   get template() { return template(posterHTML) }
-  get shouldRender() { return this.container.playback.name !== 'html_img'}
+  get shouldRender() {
+    const showForNoOp = !!(this.options.poster && this.options.poster.showForNoOp)
```
tjenkinson: I'm not sure if maybe it should be
```javascript
typeof(this.options.poster) === "object" && this.options.poster.showForNoOp
```
to be a bit more explicit that it's the object version of the setting?

src/playbacks/no_op/no_op.js
NicholasAsimov: One of the reasons for this PR is that to stop NoOp animations when it's not shown (e.g. autoPlay is false). Currently it eats up 7-10% CPU on i5 Haswell processor and it's only 1 instance of the player. If user would go on a page with multiple intances of Clappr and will get NoOp playback it will eat up like 30% CPU which is definitely not good. What can we do about it?

NicholasAsimov:
@leandromoreira Definitely will do, together with tests, after we finalize this.

src/playbacks/no_op/no_op.js
tjenkinson: I didn't realise it used up that much. Maybe we could make the noise generation more efficient?

leandromoreira:
Lint is crying :(

```
/home/travis/build/clappr/clappr/src/playbacks/no_op/no_op.js
  28:83  error  Extra semicolon  semi
```

src/plugins/poster/poster.js
towerz: I think there's no need to explicitly check if `this.options.poster` is an object. 

LGTM

PR #1174, opened 2016-10-06
Fix hls.js playback destroy.

kslimani:
Hi,

this fix an issue where hls.js instance is not properly destroyed if you destroy Clappr player while still playing video.

Currently, if you destroy Clappr player while playing, the hls.js still work and download http chunks (until the JS engine eventually garbage it).

Also i am ensure about using "delete" to nullify the instance (see [here](https://github.com/clappr/clappr/blob/master/src/playbacks/hls/hls.js#L336)).


paunadeu:
hallelujah!

leandromoreira:
Thank you very much @kslimani

PR #1188, opened 2016-10-21
Remove 'browser' field from package.json file.

### LINKED ISSUE #1171: Problem with Browserify
Hello,
I think there is an issue with the [browser](https://github.com/clappr/clappr/blob/master/package.json#L6) field in package.json file.

it should not point to minified version of Clappr. In fact, this field should be removed from package.json because the "main" field is already pointing to a "browser compatible" version of the module.

If you do a require('clappr') and use Browserify to create a bundle js file, with the current package.json it will write the minified version of Clappr into the bundle, which, i think, is not expected. (ie: it should write the non-minified version of Clappr module).

I think the "browser" field should be used only if the main field entry point is not compatible with browsers.

But i may be completly wrong here and it would be nice if any JS specialist could confirm that ? @leandromoreira ?

kslimani:
This should  fixes #1171

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,6 @@
   "version": "0.2.63",
   "description": "An extensible media player for the web",
   "main": "./dist/clappr.js",
-  "browser": "./dist/clappr.min.js",
   "scripts": {
     "release": "node_modules/.bin/webpack --progress -d --optimize-mangle --optimize-dedupe --optimize-occurence-order --output-filename clappr.min.js",
     "build": "node_modules/.bin/webpack --progress",


leandromoreira:
Let's wait for someone else to review it.

leandromoreira:
@towerz are you okay with this? You can follow the discussion [here](https://github.com/clappr/clappr/issues/1171)

towerz:
@leandromoreira seems ok to me. It actually seems to be reminiscent of our previous browserify configuration, before we changed to webpack.

Issue #1143, opened 2016-09-09
Supports HLSJS > 0.6.x-x


leandromoreira:
I tested with 0.6.2-6 and saw some errors.

update: @tjenkinson pointed out the solution at https://github.com/clappr/clappr/issues/1143#issuecomment-246997927

we're also not firing bufferfull for mp4.

paunadeu:
https://github.com/dailymotion/hls.js/releases/tag/v0.6.2-6
fix interfragment audio glitches and potential audio/video out of sync issue

We have this issue, have any ETA to solve this errors? Can you give more info about those errors?

Regards

leandromoreira:
@paunadeu I made a black box test and Clappr didn't play any video at all.

paunadeu:
Can you upload for analyze it?

Regards,
Pau

leandromoreira:
@paunadeu you can do it too:

1. `git clone https://github.com/clappr/clappr.git`
2. `cd clappr`
3. `vim package.json` -> `"hls.js": "^0.6.2-6"`
4. `npm install`
5. `npm start`

paunadeu:
Works fine on:
https://tv.fibracat.cat/cdn/clappr2/dist/

But when I use in other page don't work. I see that use .min version on this page, change to non-minified version and works properly. 

Can you check if works without minified version?

Regards,
Pau

paunadeu:
Works with minified version and 0.6.2-6 now. I did nothing.

Regards

leandromoreira:
I tried again and it shows some errors for hls:

```
Uncaught ReferenceError: _typeof2 is not defined
clappr.js:2801 [error][hlsjs: trying to recover from error, evt hlsError, data [object Object] ]
```

paunadeu:
What video are you playing? What is URI?

Regards,
Pau

2016-09-13 19:57 GMT+02:00 Leandro Moreira <notifications@github.com>:

leandromoreira:
@paunadeu http://www.streambox.fr/playlists/x36xhzz/x36xhzz.m3u8

paunadeu:
The same URL has problems for me too. 
Can u try with http://vevoplaylist-live.hls.adaptive.level3.net/vevo/ch1/appleman.m3u8

Regards,
Pau

leandromoreira:
@paunadeu the point is the same url works when we use hls.js alone http://dailymotion.github.io/hls.js/demo/ . It seems that all the headers are working fine too:

``` diff
[...]

PR #1200, opened 2016-11-03
adds support for hls.6.x

leandromoreira:
To fix https://github.com/clappr/clappr/pull/1191#event-845886043 and https://github.com/clappr/clappr/issues/1143#issuecomment-246997927

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: node_js
 sudo: false
 node_js:
-  - "5"
+  - "7"
 
 before_install:
   - export DISPLAY=:99.0
--- a/package.json
+++ b/package.json
@@ -68,7 +68,7 @@
     "mkdirp": "^0.5.1",
     "mocha": "^2.3.4",
     "node-bourbon": "^4.2.8",
-    "node-sass": "^3.7.0",
+    "node-sass": "^3.11.0",
     "s3": "^4.1.1",
     "sass-loader": "^4.0.0",
     "sinon": "^1.10.2",
--- a/package.json
+++ b/package.json
@@ -68,7 +68,7 @@
     "mkdirp": "^0.5.1",
     "mocha": "^2.3.4",
     "node-bourbon": "^4.2.8",
-    "node-sass": "^3.11.0",
+    "node-sass": "^3.11.1",
     "s3": "^4.1.1",
     "sass-loader": "^4.0.0",
     "sinon": "^1.10.2",
--- a/package.json
+++ b/package.json
@@ -80,7 +80,6 @@
     "url-loader": "^0.5.6",
     "webpack": "^1.12.1",
     "webpack-dev-server": "^1.14.0",
-    "webworkify-webpack": "1.1.4",
     "yargs": "^4.1.0"
   },
   "dependencies": {}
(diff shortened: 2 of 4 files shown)


leandromoreira:
Although the build says it's broken bc of lint I can't reproduce it locally.

```
$ npm run lint

> clappr@0.2.63 lint /Users/leandro.moreira/src/clappr
> eslint src/ test/
```

kslimani:
It is hls.js line 358 : https://github.com/clappr/clappr/blob/master/src/playbacks/hls/hls.js#L358

leandromoreira:
@kslimani if you try to run it locally it doesn't show the lint error ;(

kslimani:
it should be 

```javascript
  _fillLevels() {
    this._levels = this._hls.levels.map((level, index) => {
      return {id: index, level: level, label: `${level.bitrate/1000}Kbps`}
    })
    this.trigger(Events.PLAYBACK_LEVELS_AVAILABLE, this._levels)
  }
```

kslimani:
Locally, i have the following message

```
/home/kslimani/projects/clappr/src/playbacks/hls/hls.js
  358:5  error  Expected indentation of 6 spaces but found 4  indent

✖ 1 problem (1 error, 0 warnings)
```

kslimani:
Additionally, if i try to build Clappr with current master, hls.js playback doesn't work anymore. (but maybe you are already aware of issue ?). I am still digging where is the problem, if i findout something i will update.

leandromoreira:
@kslimani can you test this on your side? I already ran some tests (mp4, hls).

kslimani:
@leandromoreira yes i am gonna test and feedback.

kslimani:
@leandromoreira so far it is working again (testing with "npm run start" build demo page). :+1:

leandromoreira:
thank you very much @kslimani

leandromoreira:
> I think some people might be setting hlsjs options and accessing the hlsjs instance directly and this might cause problems for them?

I think these people are relying on unstable abstraction :P I'm in favor of keeping the current minor version.

shacharz:
> unstable abstraction

What abstraction are you referring to?

leandromoreira:
> What abstraction are you referring to?

@shacharz I was referring to the people that use the private instance of `hls` (`clapprInstance.core.getCurrentPlayback()._hls`) as something stable

shaharmor:
@leandromoreira what about people that used a custom XHR loader for hls.js that was passed using Clappr's config? This update broke it for them because hls.js 0.6.x has a different API for its xhr loader.

leandromoreira:
@shaharmor that's something we need to fix :) btw feel free to do a PR!

shacharz:
@leandromoreira I don't mind doing a PR, but we need to agree about it first :)
The 2 solutions I see:
1. revert to previous hls.js
2. release a major version and remove 0.2.64

leandromoreira:
oh so the new HLS doesn't offer `XHR loader` config?

shacharz:
@leandromoreira It does, but the API between the XHR loader and the rest of the library has changed

leandromoreira:
So it might be better to offer this new API or if possible to adapt both new and old clients.

Issue #863, opened 2016-03-01
Provide accessibility 


leandromoreira:
This is a huge topic, but we should at least provide minimal support for: **play**, **pause** and **mute**  we can start with the basic [aria-label](https://www.w3.org/TR/wai-aria/states_and_properties#aria-label) 

Thanks to @bernardocamilo

leandromoreira:
For the SVGS https://www.sitepoint.com/tips-accessible-svg/

leandromoreira:
An real example https://www.paypal-engineering.com/2014/09/05/introducing-an-accessible-html5-video-player/

dovematrix:
Has there been any progress on this issue?  The WAVE extension, http://wave.webaim.org/extension/, for the Google Chrome browser reports three accessibility errors.  The play/pause, full screen, and another button does not have descriptive text.

leandromoreira:
@dovematrix unfortunately, I don't think so 😞

cinnabot:
Hi, I have some suggestions to help solving this issue!



Here is a list of similar issues:


1. [Update HLS Provider used by HLS plugin](https://github.com/clappr/clappr/issues/54)

2. [HD label is not working](https://github.com/clappr/clappr/issues/130)

3. [Error with hlsjs when providing source in load method instead of options object](https://github.com/clappr/clappr/issues/589)

4. [Chromeless player should not play/pause when clicking over the video](https://github.com/clappr/clappr/issues/290)

5. [Implement the new UI](https://github.com/clappr/clappr/issues/132)






Hope this is useful!

Cheers,
Cinnabot.


PS: Liked my suggestions? Install me in your repository! Visit: [getcinnamon.io](https://www.getcinnamon.io).

bikegriffith:
Just submitted a PR to at least appease WAVE with some simple aria-labels

PR #1204, opened 2016-11-04
Provide aria labels for buttons

bikegriffith:
The most basic thing possible to start to appease #863

### DIFF
--- a/src/plugins/dvr_controls/public/index.html
+++ b/src/plugins/dvr_controls/public/index.html
@@ -1,2 +1,2 @@
 <div class="live-info"><%= live %></div>
-<button type="button" class="live-button"><%= backToLive %></button>
+<button type="button" class="live-button" aria-label="Back to live"><%= backToLive %></button>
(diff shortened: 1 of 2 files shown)


src/plugins/dvr_controls/public/index.html
```diff
@@ -1,2 +1,2 @@
 <div class="live-info"><%= live %></div>
-<button type="button" class="live-button"><%= backToLive %></button>
+<button type="button" class="live-button" aria-label="Back to live"><%= backToLive %></button>
```
leandromoreira: The names are i18n, you shouldn't replace them.

leandromoreira:
Thanks @bikegriffith I'd also like to point that one test stop working  https://travis-ci.org/clappr/clappr/builds/173304206

```
Chromium 37.0.2062 (Ubuntu 0.0.0) UIObject selects elements within the component FAILED
	Error: timeout of 2000ms exceeded. Ensure the done() callback is being called in this test.
Chromium 37.0.2062 (Ubuntu 0.0.0): Executed 50 of 179 (1 FAILED) (0 secs / 18.463 secs)
```

src/plugins/dvr_controls/public/index.html
towerz: Agree, instead of using a literal string "Back to Live", you should use the value of `backToLive`.

towerz:
Thanks @bikegriffith! 🍻

Issue #1221, opened 2016-11-23
FlasHLS playback: player.getCurrentTime() is not consistent with events


GuilhermeMedeiros:
**Browser**: 54.0.2840.98 (64-bit)

**OS**: OS X El Capitan 10.11.6

**Clappr Version**: 0.2.64

**Steps to reproduce**:

* Set the source to a HLS LIVE stream with DVR
* Use the FlasHLS playback: `plugins: [Clappr.FlasHLS]`
* Set the debug level to 0 `Clappr.Log.setLevel(0)`
* While you are watching live, compare the `player.getCurrentTime()` with the debug events

cinnabot:
Hi, I have some suggestions to help solving this issue!



Here is a list of similar issues:


1. [[FlasHLS] missing play event after starting paused LIVE stream](https://github.com/clappr/clappr/issues/901)

2. [What is the proper way to force FlasHLS?](https://github.com/clappr/clappr/issues/762)

3. [[Clappr.Log] set log level to disabled](https://github.com/clappr/clappr/issues/540)

4. [How to detect FlasHLS?](https://github.com/clappr/clappr/issues/935)

5. [[flasHLS] JFYI - flashls v0.4.4.22 is here](https://github.com/clappr/clappr/issues/1054)






Hope this is useful!

Cheers,
Cinnabot.


PS: Liked my suggestions? Install me in your repository! Visit: [getcinnamon.io](https://www.getcinnamon.io).

GuilhermeMedeiros:
```js
var playerElement = document.getElementById("player-wrapper");

Clappr.Log.setLevel(0)

window.player = new Clappr.Player({
  source: '[A LIVE HLS URL]',
  baseUrl: '/latest',
  poster: 'http://clappr.io/poster.png',
  mute: true,
  autoPlay: true,
  plugins: [Clappr.FlasHLS],
  height: 360,
  width: 640,
  events: {
    onTimeUpdate: console.log
  }
});

player.attachTo(playerElement);
```

leandromoreira:
I tried to reproduce this using one internal live stream and the results were consistent:

With a DVR of 590s the logger shows consistently 590 ... 590 forever and when we start to watch a previous moment (DVR mode) the position was consistent with duration too. What values were you expecting to see and what values did you get?

GuilhermeMedeiros:
If you are watching the DVR, then yes. But if you are watching the LIVE it will return the duration. Check the PR.

[...]

PR #1222, opened 2016-11-23
Do not verify if DVR is in use to determine current position

GuilhermeMedeiros:
Fixes https://github.com/clappr/clappr/issues/1221


leandromoreira:
@GuilhermeMedeiros we'll take a look later, thanks! 👏

leandromoreira:
We'll still look on that later @GuilhermeMedeiros (it kinda requires more attention of us)

PR #1224, opened 2016-11-29
fix wrong order triggering for Flash Playback

kafeltz:
This bug occurs when using Flash Playback was causing wrong data for report using _ClapprStats_ plugin.

The issue is the player saying the video is playing when actually is not yet, thus making the report begin to count before it should do.

The **startup** time is less than the initial **buffering**, which does not make sense. Because **startup** time began to count when play was triggered, but not when actually the video really started to play.

This log was printed by using this callback:

```
clapprStats: {
    runEach: 5000,
    onReport: function(metrics) {
        console.warn(metrics['timers'])
    },
},
```

Here the **startup** timer is bigger than **buffering**, which makes sense, after the first buffering was completed the video really **started**.
```
 Object <--- bug fixed
     buffering: 1577.9450000000006 <--- here
     latency: 0
     pause: 0
     session: 5000.075000000001
     startup: 1675.6250000000005    <--- and here
     watch: 3127.865000000007
```
And here is the wrong version logged.
```
Object <--- with bug
    buffering: 2042.39 <--- here
    latency: 0
    pause: 0i
    session: 4999.985000000001
    startup: 140.76499999999987 <--- and here
    watch: 4711.104999999999
```


kafeltz:
This line has an issue: https://github.com/clappr/clappr/blob/master/src/playbacks/flash/flash.js#L115

There is a idle state checking with a immediately empty return statement with no handling.
It should set some state and send a trigger as all other statements. 

The same issue happen here: https://github.com/clappr/clappr/blob/master/src/playbacks/flash/Player.as
There are a lot of: `playbackState` without  `_triggerEvent('statechanged');`

**Briefing:**
The correct behavior should be the _lower level component_ report its state/behavior _to upper level component_. As result, some statements are handled where it shouldn't be as the `src/playbacks/flash/flash.js`

leandromoreira:
Thanks, @kafeltz :)

I saw that there is a small  code style making the build to refuse to pass, https://travis-ci.org/clappr/clappr/builds/185463271 .

leandromoreira:
Besides the small code style thing, everything is looking great! I'll leave the merge for @towerz , I don't know if we need to change anything else to build a new flash version.

PR #1232, opened 2016-12-11
Allow to pass a custom background to poster plugin

### LINKED ISSUE #1231: Allow to give poster plugin a color instead of an image
It would be great to have an option to be able to give poster not only an image to use but a custom color as well. 

I will probably send a pull request for this if you think this is a good idea.

NicholasAsimov:
Adds an ability to specify a custom background for the poster plugin so people can not only specify an image, but do all kind of stuff like patterns, gradients or plain colors. I think it would be a good addition.

I guessed that I don't need to check if `poster.custom` is a string (it's implied) because there's no such checks for `poster` or `poster.url`.

Closes #1231

### DIFF
--- a/src/plugins/poster/poster.js
+++ b/src/plugins/poster/poster.js
@@ -116,13 +116,14 @@ export default class PosterPlugin extends UIContainerPlugin {
     let style = Styler.getStyleFor(posterStyle, {baseUrl: this.options.baseUrl})
     this.$el.html(this.template())
     this.$el.append(style)
-    if (this.options.poster) {
-      if (this.options.poster.custom) {
-        this.$el.css({'background': this.options.poster.custom})
-      } else {
-        const posterUrl = this.options.poster.url || this.options.poster
-        this.$el.css({'background-image': 'url(' + posterUrl + ')'})
-      }
+
+    const isRegularPoster = this.options.poster && this.options.poster.custom == undefined
+
+    if (isRegularPoster) {
+      const posterUrl = this.options.poster.url || this.options.poster
+      this.$el.css({'background-image': 'url(' + posterUrl + ')'})
+    } else if (this.options.poster) {
+      this.$el.css({'background': this.options.poster.custom})
     }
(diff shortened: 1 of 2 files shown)


src/plugins/poster/poster.js
```diff
@@ -117,8 +117,12 @@ export default class PosterPlugin extends UIContainerPlugin {
     this.$el.html(this.template())
     this.$el.append(style)
     if (this.options.poster) {
```
leandromoreira: How about:

```javascript
const isRegularPoster = this.options.poster && this.options.poster.custom == undefined

if (isRegularPoster) {
  const posterUrl = this.options.poster.url || this.options.poster
  this.$el.css({'background-image': 'url(' + posterUrl + ')'})
} else {
  this.$el.css({'background': this.options.poster.custom})
}
```

leandromoreira: Good catch
`else if (this.options.poster)`

NicholasAsimov:
Should I create tests which render poster and check if background was applied correctly?

leandromoreira:
Since you added behavior can you add tests for it?

https://github.com/clappr/clappr/blob/master/test/plugins/poster_spec.js 

Otherwise we can do that too.

NicholasAsimov:
@leandromoreira No problem, what kind of tests should I add as well? I did the rendering test already.

leandromoreira:
@NicholasAsimov my bad I didn't see you added tests! Thank you so much! :D 

I'm pretty happy with this addition, what about you @clappr/core ?

test/plugins/poster_spec.js
```diff
@@ -95,4 +95,13 @@ describe('Poster', function() {
     expect(this.poster.shouldHideOnPlay()).to.equal(false)
   })
 
+  it('renders custom background', function() {
```
leandromoreira: Cool! Now, we (clappr) also need to add a test case for `background-image` 👍 

Nice work @NicholasAsimov

PR #1254, opened 2017-01-11
Allows to automatically seek hls.js playback using 't' query parameter

### LINKED ISSUE #1246: Player doesn't seek with 't' query parameter
**Browser**: Chromium 55

**OS**: Linux 4.8

**Clappr Version**: 0.2.64

**Steps to reproduce**:

* Start the player with ?t=10 in the url
* Player doesn't seek to 10 seconds

Did you try to reproduce this issue at http://cdn.clappr.io/ Yes

It did work before, but I don't remember the exact version

NicholasAsimov:
This PR makes hls.js playback behave like html5_video and flashls playbacks.

It's a bit out of place I think, where would be more suitable to put this code?

Closes #1246

### DIFF
--- a/src/playbacks/hls/hls.js
+++ b/src/playbacks/hls/hls.js
@@ -8,7 +8,7 @@ import isEqual from 'lodash.isequal'
 import Events from 'base/events'
 import Playback from 'base/playback'
 import Browser from 'components/browser'
-import {now} from 'base/utils'
+import {now, seekStringToSeconds} from 'base/utils'
 import Log from 'plugins/log'
 
 const AUTO = -1
@@ -139,7 +139,11 @@ export default class HLS extends HTML5VideoPlayback {
   }
 
   _setupHls() {
-    this._hls = new HLSJS(this.options.playback.hlsjsConfig || {})
+    const autoSeekFromUrl = typeof(this.options.autoSeekFromUrl) === 'undefined' || this.options.autoSeekFromUrl
+    const startPosition = this.getPlaybackType() !== Playback.LIVE && autoSeekFromUrl && seekStringToSeconds();
+    const hlsjsConfig = Object.assign({}, this.options.playback.hlsjsConfig, {startPosition})
+
+    this._hls = new HLSJS(hlsjsConfig)
     this._hls.on(HLSJS.Events.MEDIA_ATTACHED, () => this._hls.loadSource(this.options.src))
     this._hls.on(HLSJS.Events.LEVEL_LOADED, (evt, data) => this._updatePlaybackType(evt, data))
     this._hls.on(HLSJS.Events.LEVEL_UPDATED, (evt, data) => this._onLevelUpdated(evt, data))
--- a/src/playbacks/hls/hls.js
+++ b/src/playbacks/hls/hls.js
@@ -10,7 +10,7 @@ import Playback from 'base/playback'
 import Browser from 'components/browser'
 import {now, seekStringToSeconds} from 'base/utils'
 import Log from 'plugins/log'
-import $ from 'clappr-zepto';
+import $ from 'clappr-zepto'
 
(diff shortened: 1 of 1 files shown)


NicholasAsimov:
Weird, I run tests locally and then pass, am I missing something?

kslimani:
@NicholasAsimov maybe modify `this.options.playback.hlsjsConfig` value in constructor instead of _setupHls() ?  (just a thought, did not test)

NicholasAsimov:
My tests pass locally for some reason but they complain about missing .env, what variables should be there?

kslimani:
For missing .env file, try to create an empty one locally : `touch .env` (it seems it is required by karma config file) It seems it check for TRAVIS key, if not set, it run tests using Firefox only (which may explains why it works for you locally, but fail on travis ci which use a profile with 3 browsers)

NicholasAsimov:
Seems like Object.assign requires a polyfill, that's why it fails in Chromium 37 and IE11, what to do?

kslimani:
> Seems like Object.assign requires a polyfill, that's why it fails in Chromium 37 and IE11, what to do?

don't use it 😄

NicholasAsimov:
@leandromoreira unapprove this please. This currently has problems with Live with DVR playlist. I will work on a fix later.

NicholasAsimov:
Alright, so the problem is that, obviously, in constructor playbackType will be 'vod' because it's the default value, and we don't know true playback type until the first `LEVEL_LOADED` event. But we need to specify startPosition in the hls.js contructor if I'm not mistaken. Any ideas?

src/playbacks/hls/hls.js
```diff
@@ -135,11 +136,16 @@ export default class HLS extends HTML5VideoPlayback {
     // #EXT-X-PLAYLIST-TYPE
     this._playlistType = null
     this._recoverAttemptsRemaining = this.options.hlsRecoverAttempts || 16
+
+    const autoSeekFromUrl = typeof(this.options.autoSeekFromUrl) === 'undefined' || this.options.autoSeekFromUrl
+    const startPosition = this.getPlaybackType() !== Playback.LIVE && autoSeekFromUrl && seekStringToSeconds()
```
tjenkinson: Currently this could get set to false
(Ie if type LIVE)  when I think it should be undefined

NicholasAsimov:
I wouldn't say it broke, just now it expects not a string from which to parse (used to be that so you pass url for example) and now it uses url by default and asks which query parameter to use (default is `t`)

NicholasAsimov:
I changed the code so instead of passing initial option to hlsjs I just do what you did in html5_video playback - when we get the initial metadata (so we know what type of a playlist this is) we parse the time and manually seek the player.

NicholasAsimov:
Also to fix the html5_video argument for seekStringToSeconds should be removed because it expects query parameter not a string. And it probably would make sense to rename it to something like `getSeekTimeFromUrl`.
https://github.com/clappr/clappr/blob/master/src/playbacks/html5_video/html5_video.js#L340

NicholasAsimov:
Correct

towerz:
Just noticed this has been implemented on #1334 with a simpler solution. There's no need for merging this PR.

Closing for now, but if something else comes up we can revisit this solution.

NicholasAsimov:
@towerz #1334 only fixes html5_video playback, hls.js playback still doesn't have autoseek.

kslimani:
@NicholasAsimov does it work for you with `html5_video` playback ? (just to be sure)

NicholasAsimov:
@kslimani My apologies, it actually does work. For some reason I thought that cdn.clappr.io holds the latest master instead of a last release so I tested there before I written my comment. Now I tested it with latest master and it appears that seeking is working on hls.js as well. Thanks for a great pull request!

PR #1258, opened 2017-01-12
Implement high definition check for html5_video playback

NicholasAsimov:
This aligns html5_video playback with hls.js and flashls.

### DIFF
--- a/src/playbacks/html5_video/html5_video.js
+++ b/src/playbacks/html5_video/html5_video.js
@@ -140,10 +140,17 @@ export default class HTML5Video extends Playback {
     this._handleBufferingEvents()
     this.trigger(Events.PLAYBACK_LOADEDMETADATA, {duration: e.target.duration, data: e})
     this._updateSettings()
+    this._checkHighDefinition(e)
     const autoSeekFromUrl = typeof(this._options.autoSeekFromUrl) === 'undefined' || this._options.autoSeekFromUrl
     if (this.getPlaybackType() !== Playback.LIVE && autoSeekFromUrl) {
       this._checkInitialSeek()
     }
+  }
+
+  _checkHighDefinition(e) {
+    const currentHeight = e.target.videoHeight
+    this.highDefinition = currentHeight >= 720
+    this.trigger(Events.PLAYBACK_HIGHDEFINITIONUPDATE, this.highDefinition)
   }
 
   _onDurationChange() {


src/playbacks/html5_video/html5_video.js
```diff
@@ -140,12 +140,19 @@ export default class HTML5Video extends Playback {
...
     this._updateSettings()
+    this._checkHighDefinition(e)
     const autoSeekFromUrl = typeof(this._options.autoSeekFromUrl) === 'undefined' || this._options.autoSeekFromUrl
     if (this.getPlaybackType() !== Playback.LIVE && autoSeekFromUrl) {
       this._checkInitialSeek()
     }
   }
 
+  _checkHighDefinition(e) {
+    const currentHeight = e.target.videoHeight
```
tjenkinson: I'd also call this method from the `_onTimeUpdate()` method, because with the hls playback the resolution could change at any point during a stream, if it's an adaptive stream, and if other playbacks extend this they may do the same.

src/playbacks/html5_video/html5_video.js
```diff
@@ -140,12 +140,19 @@ export default class HTML5Video extends Playback {
...
+    this.highDefinition = currentHeight >= 720
+    this.trigger(Events.PLAYBACK_HIGHDEFINITIONUPDATE, this.highDefinition)
```
tjenkinson: Should we check if the state has actually changed, and only fire the event if it has, to prevent duplicate events?

src/playbacks/html5_video/html5_video.js
```diff
@@ -140,12 +140,19 @@ export default class HTML5Video extends Playback {
...
+    this._checkHighDefinition(e)
     const autoSeekFromUrl = typeof(this._options.autoSeekFromUrl) === 'undefined' || this._options.autoSeekFromUrl
     if (this.getPlaybackType() !== Playback.LIVE && autoSeekFromUrl) {
       this._checkInitialSeek()
     }
   }
 
+  _checkHighDefinition(e) {
+    const currentHeight = e.target.videoHeight
+    this.highDefinition = currentHeight >= 720
```
tjenkinson: Also I'd change this to `this._highDefinition` as it's private, and initialise it to `false` in the constructor.

NicholasAsimov: Sure, I was looking at the hls.js playback, if we change it here should we later do the same in other playbacks? Same for the checking for duplicate events

NicholasAsimov:
@tjenkinson thanks for great feedback!

src/playbacks/html5_video/html5_video.js
NicholasAsimov: I would argue that other playbacks would need to do their own different logic because we already have different approaches in hls and html5_video. In hls js HD check happens on level switch, in html5_video it can only happen when new source loads. So by moving it to timeUpdate we're increasing number of checks that can never really change in the context of this playback.

src/playbacks/html5_video/html5_video.js
tjenkinson: Ah I didn't realise the HD event was already implemented in the hls playback. This makes it a bit more complex as this needs to not conflict with that one. I.e the hls playback condition also checks the bitrate whereas here it doesn't.

Maybe the new method `_checkHighDefinition()` could also be implemented (overridden) in the hls playback and that would be where the event is fired there too?

tjenkinson: And it might need to stay as a property 'highDefinition' as I'm not sure if code in other places expects it to be public.

NicholasAsimov: That would be a good idea to have a base _checkHighDefinition method that other playbacks would override.

NicholasAsimov:
So in order to unify

PR #1260, opened 2017-01-12
Add Mobile iPhone/AndroidChrome AutoPlay functionality

tomas2387:
You can't programmatically play a video in a mobile device.   

`Uncaught (in promise) DOMException: play() can only be initiated by a user gesture.`

The video can only be played with a user gesture or with the autoplay+muted attributes set.

Autoplay working in:
- iOS+iPhone
- Android+Firefox

Autoplay not working in:
- Android+Chrome

This feature is very useful in HLS streaming. Maybe not very useful on VOD (because the video starts muted and the user has to restart the seekbar and unmute the video :/ )


tomas2387:
Sorry for the test coverage, but I can't force `Clappr.Browser.isMobile` to true :(

leandromoreira:
@tomas2387 isn't the current clappr, autoplay and mute not working?

kslimani:
@leandromoreira : the playback `playInline` feature is now part of Clappr. (ref: #1317)

Also, according [this article](https://webkit.org/blog/6784/new-video-policies-for-ios/), it seems it is an iOS >= 10 feature only (Webkit only ?).

I think this PR is currently incomplete, because it require also to modify Clappr mute internal mechanisms to add or remove the muted property if current playback element is an HTML5 video element and Browser is using webkit with iOS >= 10.

Please note i have not yet checked if other mobile devices browser/engine policies have also introduced these changes.

leandromoreira:
@kslimani COOL, this is indeed, pretty cool! :)

tomas2387:
Thank you so much @kslimani for explaining it so well. 

I don't know what else to add to this pull request, also, I don't know how to add more coverage.
I couldn't force `Clappr.Browser.isMobile` to be true in the tests.

leandromoreira:
Thank you @tomas2387 I'll try to make a review later with a peer.

kslimani:
@tomas2387 if i understand theses changes, it automatically add the muted property to html5 video element if autoplay player option is set to true with mobile devices ?

I am not sure it is the proper way to introduce this feature. Also the core mediacontrol must be "aware" that playback is muted.

IMO, the current autoplay mechanism does not need to be changed.

Instead i would change the code parts which handle the mute (_or muted ?_) player option to properly add the `muted` element property (_on html5 video playback scenario only_). Theses changes may also apply to desktop (_i think it is acceptable also on desktop to add the muted video element property if user mute the playback_).

What do you think ?

tomas2387:
@kslimani your knowledge of clappr internals is better than mine without a doubt. And I agree with your solution. 

But:
> IMO, the current autoplay mechanism does not need to be changed.

The problem with the current autoplay mechanism is that it is doing `this.play()` without an user gesture in mobile devices (at least, that was what I saw when I created this pull request) 
https://github.com/clappr/clappr/blob/master/src/playbacks/html5_video/html5_video.js#L133

And any solution must avoid doing a `this.play()` without an user gesture in a mobile device because if we don't do that we get an ugly error:
`Uncaught (in promise) DOMException: play() can only be initiated by a user gesture.`


Or ... we can ... just ignore the error ... ?

kslimani:
@tomas2387 for example, i would remove the current modifications and try something like :

in the html5 video playback [mute](https://github.com/clappr/clappr/blob/master/src/playbacks/html5_video/html5_video.js#L228) method append this line :
```javascript
this.$el.attr({muted: 'true'})
```
 Then in the [unmute](https://github.com/clappr/clappr/blob/master/src/playbacks/html5_video/html5_video.js#L232) method append this line :
```javascript
this.$el.attr({muted: null})
```

Then try the results on mobile. (_i did not try myself, just some thoughts_).

leandromoreira:
> `Or ... we can ... just ignore the error ... ?`

This is a warning 👍  I think it's okay !

kslimani:
@tomas2387 ok here a solution, just modify the [volume](https://github.com/clappr/clappr/blob/master/src/playbacks/html5_video/html5_video.js#L224) method to this :

```javascript
  volume(value) {
    if (value === 0) {
      this.$el.attr({muted: 'true'})
    } else if (this.el.volume === 0) {
      this.$el.attr({muted: null})
    }
    this.el.volume = value / 100
  }
```

It should resolve the issue with mobile.

tomas2387:
I close this to continue with @kslimani better idea
https://github.com/clappr/clappr/pull/1432

PR #1317, opened 2017-03-09
Allows HTML5 video to play inline on iOS

kslimani:
It adds HTML5 video element attributes to play video inline in web browsers on iOS device.
I also added a new playback option to allow Apple TV AirPlay.

@towerz i failed to find any information about `x-webkit-playsinline` parameter you added when playback `playInline` is set to true. Do you have any resource link to share please ?

### DIFF
--- a/src/playbacks/html5_video/html5_video.js
+++ b/src/playbacks/html5_video/html5_video.js
@@ -113,6 +113,8 @@ export default class HTML5Video extends Playback {
       'x-webkit-playsinline': playbackConfig.playInline
     })
 
+    playbackConfig.playInline && (this.$el.attr({playsinline: 'playsinline'}))
+
     // TODO should settings be private?
     this.settings = {default: ['seekbar']}
     this.settings.left = ['playpause', 'position', 'duration']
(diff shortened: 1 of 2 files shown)


towerz:
@kslimani this attribute `x-webkit-playsinline` is the precursor for the playsinline attribute. It was documented in the official Safari/webkit docs, but the docs have been updated since then. That's why you can't find documentation on this anymore.

kslimani:
Thank you @towerz for plugin i missed this one.
I am gonna remove the AirPlay part then and only keep the prefix-less `playsinline` attribute.

kslimani:
I removed AirPlay flag and merged to single commit.

Craytor:
Do we have a timeline to when this may be implemented? Waiting for a feature like this.

kslimani:
ping @towerz 👍

towerz:
@kslimani sorry, I've been away for a while. Merging now.

NikosEfthias:
this doesnt seem to be working for ios. form me the only way out was switch to another layer from clappr

PR #1318, opened 2017-03-09
Add HTML5 video element recycle feature.

kslimani:
See #1319 for example usage and technical explanations.

This change targets mobile devices and does not change current player behaviours.

Just set the `recycleVideo` playback option to `true` to enable this feature.

It should fixes a lot of issues on mobile and tablet devices (_single tap playback source load with autoplay, advert plugins, playlist plugins, ..._)

_EDIT: merged to single commit and updated explanations._

### DIFF
--- a/src/base/ui_object.js
+++ b/src/base/ui_object.js
@@ -2,7 +2,7 @@
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 
-import {uniqueId} from './utils'
+import {uniqueId, DomRecycler} from './utils'
 import $ from 'clappr-zepto'
 import result from 'lodash.result'
 import BaseObject from './base_object'
@@ -195,7 +195,7 @@ export default class UIObject extends BaseObject {
       const attrs = $.extend({}, result(this, 'attributes'))
       if (this.id) {attrs.id = result(this, 'id')}
       if (this.className) {attrs['class'] = result(this, 'className')}
-      const $el = $('<' + result(this, 'tagName') + '>').attr(attrs)
+      const $el = DomRecycler.create(result(this, 'tagName')).attr(attrs)
       this.setElement($el, false)
     } else {
       this.setElement(result(this, 'el'), false)
--- a/src/base/utils.js
+++ b/src/base/utils.js
@@ -4,6 +4,7 @@
 /*jshint -W079 */
 
 import Browser from 'components/browser'
+import $ from 'clappr-zepto'
 
 function assign(obj, source) {
   if (source) {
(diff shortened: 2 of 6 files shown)


kslimani:
I added an new optional `recycleVideo` playback option to enable this feature.
It is disabled by default, so there is no behaviour change in player. It become dev responsability to enable or not this feature for mobile device.

kslimani:
My mistake, forgot to run tests... (working on it...)

kslimani:
The "common" bridge between changed class was `utils.js`, so my choice was to create a VideoStore static class inside in charge to store video elements indexed by player id.

@leandromoreira @tjenkinson @towerz @thiagopnts what do you think about this ?

src/base/ui_object.js
```diff
@@ -188,14 +188,33 @@ export default class UIObject extends BaseObject {
   /**
    * ensures the creation of this ui component
    * @method _ensureElement
+   * @param {Object} options the options object
    * @private
    */
-  _ensureElement() {
+  _ensureElement(options) {
```
towerz: I'd suggest not doing this in `ensureElement` directly, as it's used by every component/plugin that renders an element. Instead, since it's a feature specific to the playback implementations, I'd move this logic to either the `Playback` base class or the HTML5 video playback.

src/base/utils.js
```diff
@@ -235,10 +235,34 @@ export function removeArrayItem(arr, item) {
   }
 }
 
+// HTML5 video element recycle storage object.
+// If playback recycleVideo flag is set, it store
+// one video element for each player instance created.
+const videoStore = {}
+
+export class VideoStore {
```
towerz: Another suggestion I'd like to do is to create a stack of `video` elements instead of using `playerId` as a key since we can have more than one `video` element in a single player (for example, if you have more than one container).

src/base/ui_object.js
kslimani: I am not sure how technically to do that ? the base ui object _ensureElement() contains the logic for html element creation. Do you have any tips ?

src/base/utils.js
kslimani: @towerz what are the requirements for have more than one video element in single player ? _(i have not yet find any infos to reproduce this behaviour)_

kslimani: For exemple, if move the logic in `HTML5Video` constructor, i may add it just after super() call and then store video element if none available or override current one by recycled ? (in this case the parent _ensureElement() method has created a video element for nothing, but maybe it is an acceptable micro overheat ?)

kslimani: What about a Factory class which contains recycle logic, configured by core, with a `createElement(tagName)` method ?

This method always return a Zepto element with the following logic : if recycle is enabled and requested tag name is video, then check in stack if video a element is available. If one is available, remove it from stack and return it. Otherwise create a new one and return it. _(of course if requested tag name is not video, simple return new requested Zepto Element)_

Then in HTML5Video class destroy method, if recycle is enabled then put video element back to stack using a method from factory class. Also this method, could cleanup the previously added style children(s) nodes. (also cleanup attributes and classes ?)

I see no other way to perform that feature. If the logic is moved "higher" in parent playback class, the associated dom element is already created and have some events binded (_ensureElement + delegateEvents in UIObject_).

But i may be wrong and missed a proper way to perform this. I am open to any idea or hints.

PS: also if someone could give me an example case where Clappr create multiple video element (so i can test/debug it).

kslimani:
I moved all the video recycle logic into DomRecycler class in utils which now uses a stack.
It should work event if multiple videos are created in player.

Changes goal is to allow Clappr player to behave like [this page](http://www.dailymotion.com) with mobile devices. (_ie: click to start video, then single click on thumbnail to start video without requiring additional click_)

This behaviour is not possible with current version of player.

leandromoreira:
Cool, thank you @kslimani [we were also looking/working](https://github.com/clappr/clappr/projects/1) with something to solve this problem : )

kslimani:
@leandromoreira what do you think about to current proposition ?

Also i would like to ensure it is properly working with scenario where several video elements are created, but i dontt know how to produce this behaviour. Do you have some hints/infos ? (_an external plugin i missed ?_)

kslimani:
I merged changes into a single commit.

kslimani:
👍 yep, it is the more "elegant" way i found with current implementation.

PR #1335, opened 2017-03-17
Fix 404 http error with unexpected poster value.

### LINKED ISSUE #1330: Minor bug in HTML5 playback poster option
In [HTML5Playback](https://github.com/clappr/clappr/blob/master/src/playbacks/html5_video/html5_video.js#L109) directly extending poster attribute with `this.options.poster` value may lead to 404 error.

This is because `this.options.poster` may be an object (_javascript will convert object to string and try to load non-existing asset_).

A fix would be to first ensure that `this.options.poster` is a string, or to "normalize" poster option in upstream code to get poster url from expected object property.

EDIT: step to reproduce : go to [demo page](http://clappr.io/demo/) and use the following javascript :

```javascript
var playerElement = document.getElementById("player-wrapper");

  poster: {url: 'http://clappr.io/poster.png'},
  mute: true,
  height: 360,
  width: 640
});

player.attachTo(playerElement);
```

(_just check on browser dev.tools network console for 404 error_)

kslimani:
Fixes #1330 .

### DIFF
--- a/src/playbacks/html5_video/html5_video.js
+++ b/src/playbacks/html5_video/html5_video.js
@@ -104,9 +104,18 @@ export default class HTML5Video extends Playback {
     const playbackConfig = this.options.playback
     const preload = playbackConfig.preload || (Browser.isSafari ? 'auto' : this.options.preload)
 
+    let posterUrl // FIXME: poster plugin should always convert poster to object with expected properties ?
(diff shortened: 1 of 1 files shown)


towerz:
Thanks @kslimani! :beers:

PR #1353, opened 2017-04-05
Fix core style element creation.

kslimani:
A bug happen when you load another source multiple times without destroying the player : it append another style element in DOM for each loaded source.

This fix prevents it by keeping a reference and properly remove existing element each render call.


towerz:
Thanks @kslimani! 🍻

PR #1354, opened 2017-04-06
fix(container): returning event when trigger CONTAINER_CONTEXTMENU

vagnervjs:


towerz:
Nice catch, @vagnervjs!

PR #1367, opened 2017-04-21
Adds support for Closed Caption control.

itsjamie:
With initial support for loading WebVTT captions through HLS.js.


kslimani:
👍  @itsjamie can you also remove line 10 of hls.js playback class to attempt to fix Travis lint check. (ie: `import Browser from 'components/browser'` ? which is not used anymore and is why travis seems to fail)

itsjamie:
@kslimani Done.

towerz:
Thanks @itsjamie ! :beers:

PR #1374, opened 2017-04-27
Enhance hls.js playback fatal network error handling.

kslimani:
Currently, if player try to load an HLS source playlist (_using hls.js playback_) which return a 40x or 50x error, then player is stalled.

The reason is that current code try to recover using `hls.startLoad()` on hls `NETWORK_ERROR`.
But according [documentation](https://github.com/video-dev/hls.js/blob/master/doc/design.md#error-detection-and-handling), the following network errors (_ErrorDetails_) cannot be recovered with `hls.startLoad()` : `MANIFEST_LOAD_ERROR`, `MANIFEST_LOAD_TIMEOUT` and `MANIFEST_PARSING_ERROR`.

This fix trigger a `PLAYBACK_ERROR` event to allow developer to decide how to handle the error. (_for example, using player `onError` configuration option_).

Usage example :

```javascript
  source: 'http://some.bad/path/playlist.m3u8',
  events: {
    onError: function(e) {
      // Here decide what to do
      // player.load('http://another.good/fallback/playlist.m3u8', null, true);
    }
  }
});
```

### DIFF
--- a/src/playbacks/hls/hls.js
+++ b/src/playbacks/hls/hls.js
@@ -250,7 +250,7 @@ export default class HLS extends HTML5VideoPlayback {
           switch(data.details) {
           // The following network errors cannot be recovered with HLS.startLoad()
           // For more details, see https://github.com/video-dev/hls.js/blob/master/doc/design.md#error-detection-and-handling
-          // Note that level load error & timeout are fatal ONLY when media is not buffered and therefore unrecoverable. This happen with dynamic manifest and stream not yet available.
+          // For "level load" fatal errors, see https://github.com/video-dev/hls.js/issues/1138
           case HLSJS.ErrorDetails.MANIFEST_LOAD_ERROR:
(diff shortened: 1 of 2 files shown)


kslimani:
Changed to "WIP", don't merge it yet , i need to validate some mechanics first with level load fatal events.

kslimani:
Added `LEVEL_LOAD_ERROR` and `LEVEL_LOAD_TIMEOUT` fatal error events which obviously cannot be automatically recovered. ([Reference issue](https://github.com/video-dev/hls.js/issues/1138))

leandromoreira:
maybe you could write a simple sample on how to deal with network error on a separated readme, what do you think @kslimani ?

kslimani:
sure, i think also i will additionally create an example plugin which satisfy [#1381](https://github.com/clappr/clappr/issues/1381#issuecomment-299153811)

PR #1377, opened 2017-04-28
Handle uncaught (in promise) DOMException with Chrome browser

kslimani:
In Google Chrome, `HTMLMediaElement.play()` returns a Promise which may trigger a DOMException when playback attempt has failed.

When it happen, it display the following error message in Google Chrome console :

```
Uncaught (in promise) DOMException: The play() request was interrupted by a call to pause().
```

This fix simply catch the exception to avoid the error message to be displayed in console.

For more details, see [Google developers article](https://developers.google.com/web/updates/2016/03/play-returns-promise).


towerz:
Thanks @kslimani!

PR #1378, opened 2017-04-28
[PROPOSITION] Expose embedded hls.js library in hls playback.

kslimani:
While working on #1374 i thinked it could be convenient to be able to access to Hls.js library for constants.

Example usage :

```javascript
import Clappr from 'clappr'

const HLSJS = Clappr.HLS.HLSJS

let player = new Clappr.Player({
  source: 'http://example.com/playlist.m3u8',
  events: {
    onError: e => {
      if (e.error && e.error.evt === HLSJS.Events.ERROR) {
        // Handle Hls.js error
      }
    }
  }
});
```


towerz:
Thanks @kslimani !

PR #1409, opened 2017-05-17
[PROPOSITION] Screen orientation detection for mobile device

kslimani:
This introduce a new core event (_only triggered on mobile devices_) when screen orientation has changed.

Motivations: It's a requirement for a feature i am working on : fullscreen on orientation change.
(_please note this PR only add screen orientation core event, not the fullscreen feature_)

Currently, fullscreen on orientation change work only on iOS devices because it is not a "true" fullscreen.
But hopefully it may be allowed in future browser versions. (_so it may be usefull for Clappr to be ready for this_).

### DIFF
--- a/src/components/core/core.js
+++ b/src/components/core/core.js
@@ -72,6 +72,7 @@ export default class Core extends UIObject {
     $(document).bind('fullscreenchange', this._boundFullscreenHandler)
     $(document).bind('MSFullscreenChange', this._boundFullscreenHandler)
     $(document).bind('mozfullscreenchange', this._boundFullscreenHandler)
+    Browser.isMobile && $(window).bind('resize', (o) => {this.handleWindowResize(o)})
   }
 
   configureDomRecycler() {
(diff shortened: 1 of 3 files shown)


src/components/core/core.js
```diff
@@ -203,6 +204,17 @@ export default class Core extends UIObject {
...
   }
 
+  handleWindowResize(event) {
+    let orientation = ($(window).width() > $(window).height()) ? 'landscape' : 'portrait'
+    if (this._screenOrientation === orientation) return
+    this._screenOrientation = orientation
+
+    this.trigger(Events.CORE_SCREEN_ORIENTATION_CHANGED, {
+      event: event,
+      orientation: this._screenOrientation
```
leandromoreira: I'm in favor of using an `Object` here even if it has a single element now `{orientation: 'x'}` and then change it also in doc (Event) because it leaves room for new attributes, what do you think?

leandromoreira: What do you think too @clappr/core ?

kslimani: @leandromoreira it is already an object with two properties : `event` which is the window resive event and `orientation` which is a string which equals 'landscape' or 'portrait'.

kslimani: Basically, if you listen to the event, you will receive an object like :
```javascript
{
  orientation: 'landscape',
  event: {
    isTrusted: true,
    // [...]
  }
}
```

leandromoreira: You're right, my bad!

kslimani:
Also added a small fix for Browser : currently on iOS device, `Browser.isSafari` return `true` with Chrome browser. This fix ensure that `Browser.isSafari` properly return `false` with Chrome Browser.

It is usefull for use case where you specifically needs to detect Safari. (_Please note i am also aware that technically on iOS Chrome browser is Safari with extended capabilities._)

_PS: if required, i can move Browser modifications on a separate PR._

towerz:
Thanks, @kslimani! 🍻

leandromoreira:
@kslimani it's even live on docs http://clappr.github.io/classes/Events.html#event_CORE_CONTAINERS_CREATED

PR #1413, opened 2017-05-19
fix(utils): allow iOS device to exit fullscreen programmatically

kslimani:
On iOS device (_playing HTML5 video_) if you call `Utils.Fullscreen.cancelFullscreen()` it will fail.

This fix allow you, to exit full screen programmatically. For example :

```javascript
const video = player.core.getCurrentPlayback().el // Assume <video> element
Utils.Fullscreen.cancelFullscreen(video)
```

It does **not** introduce any regression or breaking change.

_Tested on iPhone (Safari, Chrome), iPad (Safari, Chrome), MacOSX (Safari, Chrome, Firefox), Linux Debian (Chrome, Firefox); Window 7 (IE11, Chrome, Firefox), Windows 10 (Edge, Chrome, Firefox)._


towerz:
Thanks, @kslimani! 🍻

PR #1415, opened 2017-05-22
docs(playback): add isMobile workaround for recycleVideo playback option

kslimani:
I updated the `recycleVideo` playback option documentation : some Firefox "desktop" browser version may sometime have an issue loading video. I recommend to use this option only with mobile devices.

### DIFF
--- a/doc/BUILTIN_PLUGINS.md
+++ b/doc/BUILTIN_PLUGINS.md
@@ -81,7 +81,7 @@ The configuration for the playback, it's still only compatible with `html5_video
     controls: true,
     playInline: true, // allows inline playback when running on iOS UIWebview
     crossOrigin: 'use-credentials',
-    recycleVideo: true // Recycle <video> element for mobile device. (default is false)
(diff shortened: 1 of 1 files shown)


leandromoreira:
Did you see any error while error while this happens on firefox?

kslimani:
@leandromoreira this error happen when you attempt to load another source (for example using player.load method) with Firefox "desktop".

Nothing on console except the classical `URI invalide. Le chargement de la ressource média  a échoué.` error message. (In english, it must be something like `Invalid URI. Load of media resource failed.`).

Please note this message is displayed __even__ if `recycleVideo` is left to false.

When recycle video is enabled the only difference i noted is that a __MediaError__ with code = 4 and message = "" is triggered.

Sometimes you can recover using player `onError` event, but most of time it fail.

It think it is only a Firefox issue, i was able to reproduce the issue with Linux Debian 8 and Windows 7. (_i did not test on other OS_).

So far i have not seen any issue with other browsers (Chrome, Safari); And since this feature is only usefull with mobile devices this is why i recommend this approach.

PR #1422, opened 2017-05-23
fix(loader): update removeDups() to priorize external plugins

kslimani:
This fix plugin order when using external plugins.

EDIT:

Let's assume this code :

```javascript
class Toto extends FlasHLS {
  // [...]
}
```

By adding `plugins: [Toto]` to player option, current loader removeDups() method will return :

```javascript
[ HLS(), HTML5Video(), HTML5Audio(), Flash(), Toto(), HTMLImg(), NoOp() ]
```

With fix :

```javascript
[ Toto(), HLS(), HTML5Video(), HTML5Audio(), Flash(), HTMLImg(), NoOp() ]
```

### DIFF
--- a/src/components/loader/loader.js
+++ b/src/components/loader/loader.js
@@ -78,6 +78,7 @@ export default class Loader extends BaseObject {
 
   removeDups(list) {
     const groupUp = (plugins, plugin) => {
+      plugins[plugin.prototype.name] && delete plugins[plugin.prototype.name]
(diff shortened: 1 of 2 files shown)


kslimani:
WIP: fixing tests...

kslimani:
Updated `loader_spec` tests.

towerz:
Thanks, @kslimani! 🍻

PR #1423, opened 2017-05-23
Avoid IE11 to download .swf file twice

kslimani:
This fixe remove an unused legacy query parameter in base flash playback template (_movie param_).

It avoid IE11 to download twice the swf file. (_on player creation, the swf is downloaded once and second request is aborted. On player.load() the .swf file is downloaded only once_)


towerz:
Thanks @kslimani! 🍻

PR #1431, opened 2017-06-02
browser: add Edge detection

lucasrodcosta:
Adds a detection for Microsoft Edge browsers and a new flag called `Browser.isEdge`.

Reference: https://stackoverflow.com/questions/30591706/what-is-the-user-agent-string-name-for-microsoft-edge


towerz:
Thanks @lucasrodcosta! 🍻

PR #1432, opened 2017-06-02
mute or unmute video on volume change (muted autostart on mobile devices)

kslimani:
These changes allow autoplay on mobile devices when `mute` player option is set to true. (_and playback `playInline` is set to true on iOS_).

Thank you to @tomas2387 for original pull request #1260.

Example usage :

```javascript
var player = new Clappr.Player({
  parentId: '#player-wrapper',
  source: 'http://clappr.io/highline.mp4',
  poster: 'http://clappr.io/poster.png',
  mute: true,
  autoPlay: true,
  playback: { playInline: true }, // Required only by iOS
  height: 360,
  width: 640
});
```

Requirements :

* iOS version >= 10
* Android with Chrome version >= 53

supersede #1260, should fixes #1314

PR #1433, opened 2017-06-05
feat(media_control): allow mute/unmute video with Safari on mobile

kslimani:
Because of #1432 it is now possible to display at least the mute/unmute icon on media controls with Safari browser on mobile devices.
(_for example, it allow user to unmute video if autoplayed with mute option to true_).

### DIFF
--- a/src/components/media_control/media_control.js
+++ b/src/components/media_control/media_control.js
@@ -685,8 +685,14 @@ export default class MediaControl extends UIObject {
       this.hide()
     }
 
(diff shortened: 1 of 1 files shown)


kslimani:
Do not merge yet, it work with iOS >=10 but not if lesser it seems.

kslimani:
Added a browser version check. (_mute player option has no effect with Safari iOS lesser than 10_)

Issue #1435, opened 2017-06-07
WebVTT support


alexandrevicenzi:
**Browser**: Any

**OS**: Any

**Clappr Version**: 0.2.25

**Steps to reproduce**:

* Try to load a playlist (m3u8) with subtitles

I couldn't find anywhere how to change the current subtitle track on Clappr, neither how to disable it. By default, it's loading English subtitle.

I found a [plugin](https://github.com/PreDeToR/clappr-audio-track-selector-plugin) to switch audio track, is there anything similar for subtitles? or is there any plans to support these features on Clappr without any other plugin.

Here is a good [playlist](https://bitdash-a.akamaihd.net/content/sintel/hls/playlist.m3u8) to test. This playlist has 4 audio tracks and 4 subtitles track.

This is related with #626.

kslimani:
You are correct @alexandrevicenzi i was able to reproduce this bug.

I proposed a fix for at least be able to enable/disable CC. The default subtitle is selected by hls.js (_English is default_). It seems there is no api/configurations options to change this yet. (_i may be wrong!_)

BTW thank you for the playlist, it helped me a lot !

alexandrevicenzi:
@kslimani I think hls.js doesn't have this option. Testing in [https://video-dev.github.io/hls.js/demo/](https://video-dev.github.io/hls.js/demo/) there's only options to toggle audio and quality, the CC is part of video component I think.

kslimani:
@alexandrevicenzi yes, i think too that the switch mechanisms should work directly with video element as hls.js is listening it and is aware of subtitle switch. I will take a look at switch audio plugin to see if i can do something for CC.

kslimani:
@alexandrevicenzi i updated #1440 which should now properly handle multiple CC tracks. Could you please test if these changes satisfy your issue ?

PR #1440, opened 2017-06-09
feat: closed captions internal plugin

kslimani:
Should satisfy #1435.

It seems that hls.js [has no mechanisms](https://github.com/video-dev/hls.js/issues/1184) yet to switch subtitle track, so i added one in `html5_video` playback.

Closed captions (subtitles) feature has been isolated in an `ClosedCaptions` internal plugin to be "compatible" with @thiagopnts current work.

All the CC switch mechanisms has been moved into `html5_video` playback.

BREAKING CHANGE:  renamed subtitles events (_subtitle available and subtitle changed events_).

In Playback and Container base classes i added : 
* `closedCaptionsTracks` getter which return a "normalized" array of available subtitles tracks
* `closedCaptionsTrackId` property to get or change the current enabled track index (-1 is to hide all tracks)

The plugin is heavily inspired from [level selector plugin](https://github.com/clappr/clappr-level-selector-plugin).

I tested with HLS source ([this one](https://bitdash-a.akamaihd.net/content/sintel/hls/playlist.m3u8)). (_native and hls.js playbacks_)

Added support for external tracks in html5_video playback (_webvtt if supported by browser_).

Thx again to @itsjamie for initial implementation !

Known issues: 

* on Firefox, switching track may not instantly change text language displayed on video. It seems this is due to Gecko implementation.
* CC with HLS source seems to be not currently supported by IE and Edge.

Playback external tracks usage example :

```javascript
  parentId: '#player-wrapper',
  height: 360,
  width: 640,
  playback: {
    crossOrigin: 'anonymous', // Required if track loaded from another domain
    externalTracks: [
      {lang: 'en', label: 'English', src: 'https://static.playmedia-cdn.net/clappr/en.vtt'},
      {lang: 'fr', label: 'French', src: 'https://static.playmedia-cdn.net/clappr/fr.vtt'}
    ],
  },
});
```

TODO LIST :

- [x] hls.js playback support
- [x] add CC detection in html5 video playback (to support mp4+vtt or native hls on iOS)
- [x] add track change detection (to support native device mediacontrol interactions)
- [x] add support for html5_video `<track />`
- [x] add documentation for plugin and new `externalTracks` playback option
- [x] add polyfills module (imported in utils.js) with [Array.prototype.find](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Array/find)
- [x] add test coverage (_to the extent possible_)


src/playbacks/html5_video/html5_video.js
```diff
@@ -494,25 +495,33 @@ export default class HTML5Video extends Playback {
...
-      } else if (track.track.mode === 'showing') {
+        showingTrack = track
+      }
+      if (track.track.mode !== 'hidden') {
+        // In theory 'disabled' may be more optimized, but
+        // it leads to weird behaviours with some browsers.
+        // FIXME: value set could be browser specific ?
         track.track.mode = 'hidden'
       }
     }
```
leandromoreira: Here I would try to make it only a matter of transformations, filtering and select also.

```javascript
// since trackId is a parameter and it seems to be never changed, we can check it before all
if (!isNumber(trackId) || trackId < 0) {return}

showingTrack = tracks.select((track) => {
  if (track.id ==trackId) 
     return track
})

tracks.filter((track) => {track.track.mode !== 'hidden'})
           .forEach((track) => {track.track.mode = 'hidden'})

!!showingTrack && showingTrack.track.mode = 'showing'
this._ccTrackId = trackId
this.trigger(...)
```

kslimani: thx @leandromoreira , i will apply these changes. Note also that trackId can equals -1 (_to disable cc_).

src/base/playback.js
```diff
@@ -132,6 +132,32 @@ export default class Playback extends UIObject {
...
+    return -1
+  }
+
+  /**
+   * sets the selected closed caption track index.
+   * @method setClosedCaptionsTrack
+   * @param {Number} trackId
+   * @return {Boolean} `true` on success, otherwise `false`
+   */
+  setClosedCaptionsTrack(trackId) {} // eslint-disable-line no-unused-vars
```
kslimani: @leandromoreira you are correct, (_i missed this one because in first implementation the set was returning a boolean_). will fix also this.

src/playbacks/html5_video/html5_video.js
```diff
@@ -426,6 +447,96 @@ export default class HTML5Video extends Playback {
...
+        this.handleTextTrackChange = this._handleTextTrackChange.bind(this)
+        this.el.textTracks.addEventListener('change', this.handleTextTrackChange)
+      }
+      this._ccIsSetup = true
+    }
+  }
+
+  _handleTextTrackChange() {
+    let tracks = this.closedCaptionsTracks
+    let track = tracks.find((track) => {
```
leandromoreira: Do we have `find` prototyped on Array ourselves? I think this function doesn't exist in most mobile browser =/ according to https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Array/find?v=example

leandromoreira: Also, I think you can only return `true` or `false` to this function =)

`find((track) => { track.track.mode === 'showing' })` it still will returns the selected track

src/playbacks/html5_video/html5_video.js
```diff
@@ -426,6 +447,96 @@ export default class HTML5Video extends Playback {
...
+  }
+
+  get closedCaptionsTracks() {
+    let id = 0
+    let newId = () => { return id++ }
+    let onlySubtitles = (track) => { return track.kind === 'subtitles' }
+    let makeTrack = (track) => { return {id: newId(), name: track.label, track: track} }
+    let textTracks = this.el.textTracks ? Array.from(this.el.textTracks) : []
+
+    return textTracks.filter(onlySubtitles).map(makeTrack)
```
leandromoreira: Cool!  :D `map` and `filter` seems to be guaranteed (see https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Array/map?v=example https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Array/filter?v=example)

Sorry if I was nitpicking to change the code style =/, I believe that this style is easier to maintain in long term.

src/playbacks/html5_video/html5_video.js
```diff
@@ -426,6 +447,96 @@ export default class HTML5Video extends Playback {
...
+    if (!isNumber(trackId)) {
+      return
+    }
+
+    let tracks = this.closedCaptionsTracks
+    let showingTrack
+
+    if (trackId !== -1) {
+      // Get track to show
+      showingTrack = tracks.find((track) => {
```
leandromoreira: the same logic for boolean predicated

src/playbacks/html5_video/html5_video.js
```diff
@@ -426,6 +447,96 @@ export default class HTML5Video extends Playback {
...
+    // Since it is possible to display multiple tracks,
+    // ensure that all tracks are hidden.
+    tracks
+      .filter((track) => { return track.track.mode !== 'hidden' })
+      .forEach((track) => { track.track.mode = 'hidden' })
+
+    // Display track
+    if (showingTrack) {
+      showingTrack.track.mode = 'showing'
+    } else if (trackId !== -1) {
```
leandromoreira: Since `trackId` never get changed during this function life cycle I'm in favor of moving this statement up above like if it was a [precondition](https://github.com/google/guava/wiki/PreconditionsExplained) to run this functions? what do you think? It's like, if the track is different from `-1` then we should be doing nothing, I'm positively saying this because we didn't change its value.

Should be mark all the tracks as hidden even if the trackId is different from -1?

Does `-1` means selected?

src/plugins/closed_captions/closed_captions.js
```diff
@@ -0,0 +1,133 @@
+import UICorePlugin from '../../base/ui_core_plugin'
+import template from '../../base/template'
+import Styler from '../../base/styler'
+import Events from '../../base/events'
+import ccIcon from '../../icons/09-cc.svg'
+import ccStyle from './public/closed_captions.scss'
+import ccHTML from './public/closed_captions.html'
+
+export default class ClosedCaptions extends UICorePlugin {
```
leandromoreira: 🥇

src/playbacks/html5_video/html5_video.js
kslimani: Indeed (i did not pay attention to browser support =/) it was the function which was the most approaching "select" (_but i think select is ruby array function_). I guess i will use a simple function instead (_or should i add polyfill ?_)

src/playbacks/html5_video/html5_video.js
kslimani: thx to you for your time and learning tips !

src/playbacks/html5_video/html5_video.js
kslimani: -1 means "disabled", ie: if set to -1 if hide subtitles. (_there is a "disabled" label in the select menu_).
nice catch, you are right, i can move up the condition to avoid filter and not change tracks state.

src/playbacks/html5_video/html5_video.js
kslimani: In the end I added an arrayFind method to utils module.

leandromoreira: About this decision, what name, where or if we should put it on utils or as a polyfill I think @clappr/core can say something 👀 

I think `select` or `find` should be better and as polyfill to make it flow with the existent `filter`, `map` and others.

kslimani: Up to you. i could :

* just rename to utils.find
* rename to find and use this code : https://github.com/stefanduberg/array-find/blob/master/find.js
* use the polyfill from https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Array/find?v=example#Polyfill
* something else ?

Just tell me 😄

leandromoreira: You're doing an awesome job 🍻  I'll let @clappr/core decide this one, we judge much better than only my view 😄

leandromoreira: But I'm in favor of using Mozilla polyfill (referencing them on the source too)

kslimani:
@leandromoreira : in the scenario of using Mozilla polyfill, where do you think i should put the code in Clappr ? (_since it's a polyfill, it should be run as soon as possible ?_)

leandromoreira:
@kslimani I would like to listen to @clappr/core thoughts!? I'm gonna find someone to review this too! 👀

towerz:
@leandromoreira @kslimani I believe we can create a `polyfills` module that could be imported by `utils.js`. In this module, we could just add the polyfills as needed, and it would be executed right when Clappr script is processed.

kslimani:
Added `polyfills` module and fixed an issue for iOS 9.x devices with [crossorigin](https://www.w3.org/TR/html5/embedded-content-0.html#attr-media-crossorigin) video element attribute.

src/base/polyfills.js
```diff
@@ -0,0 +1,53 @@
...
+
+/**
+ * Array.prototype.find
+ *
+ * Original source : https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Array/find
+ * See also : https://tc39.github.io/ecma262/#sec-array.prototype.find
+ */
+if (!Array.prototype.find) {
+  Object.defineProperty(Array.prototype, 'find', {
+    value: function(predicate) {
```
kslimani: btw, i am not sure if i can rewrite this line to `value: predicate => {`, using `=>` may bind function to this ? any tought ?

kslimani:
@leandromoreira @towerz : i updated code, it should be done for another review tour !

towerz:
Great job, @kslimani! 🍻

kslimani:
next step is to add CC capabilities to dash playback ! 😃

PR #1455, opened 2017-06-29
Implement touch event support in MediaControl

### LINKED ISSUE #1205: iOS native controls appear when exiting full screen
**Browser**: chrome

**OS**: iOS 10.1.1 iPhone

**Clappr Version**: 0.2.64

**Steps to reproduce**:

- clIck play
- player plays full screen
- click full screen button to exit full screen
- playback continues inline (new iOS feature)
- native iOS control bar now appears below media control.

Not sure if this is something we can fix or an iOS bug.

https://goo.gl/photos/MebkTzzftNr8BZSE7

bikegriffith:
Improves usability on mobile devices by allowing finger-based dragging of the scrubber.

This PR also fixes #1205 by explicitly removing the `controls` attribute after `video. webkitendfullscreen` is fired.

### DIFF
--- a/src/components/media_control/media_control.js
+++ b/src/components/media_control/media_control.js
@@ -45,23 +45,38 @@ export default class MediaControl extends UIObject {
 
   get events() {
     return {
+      // play/pause/stop
       'click [data-play]': 'play',
       'click [data-pause]': 'pause',
       'click [data-playpause]': 'togglePlayPause',
       'click [data-stop]': 'stop',
       'click [data-playstop]': 'togglePlayStop',
+
+      // fullscreen toggle
       'click [data-fullscreen]': 'toggleFullscreen',
-      'click .bar-container[data-seekbar]': 'seek',
+
+      // volume
       'click .bar-container[data-volume]': 'onVolumeClick',
       'click .drawer-icon[data-volume]': 'toggleMute',
-      'click [data-cc-button]': 'toggleClosedCaptions',
       'mouseenter .drawer-container[data-volume]': 'showVolumeBar',
       'mouseleave .drawer-container[data-volume]': 'hideVolumeBar',
       'mousedown .bar-container[data-volume]': 'startVolumeDrag',
       'mousemove .bar-container[data-volume]': 'mousemoveOnVolumeBar',
+
+      // closed captions
+      'click [data-cc-button]': 'toggleClosedCaptions',
+
+      // seek
+      'click .bar-container[data-seekbar]': 'seek',
       'mousedown .bar-scrubber[data-seekbar]': 'startSeekDrag',
       'mousemove .bar-container[data-seekbar]': 'mousemoveOnSeekBar',
       'mouseleave .bar-container[data-seekbar]': 'mouseleaveOnSeekBar',
+      // ... seek with touch
+      'touchstart .bar-scrubber[data-seekbar]': 'startSeekDrag',
+      'touchmove .bar-container[data-seekbar]': 'touchMoveOnSeekBar',
+      'touchend .bar-container[data-seekbar]': 'touchEndOnSeekbar',
+
+      // show/hide when mouse over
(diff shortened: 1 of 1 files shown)


src/components/media_control/media_control.js
```diff
@@ -110,26 +125,36 @@ export default class MediaControl extends UIObject {
...
-      this.listenTo(this.container, Events.CONTAINER_VOLUME, this.onVolumeChanged)
-      if (this.container.playback.el.nodeName.toLowerCase() === 'video') {
-        // wait until the metadata has loaded and then check if fullscreen on video tag is supported
-        this.listenToOnce(this.container, Events.CONTAINER_LOADEDMETADATA, this.onLoadedMetadataOnVideoTag)
-      }
+    if (!this.container) {
+      return;
+    }
+
+    Mediator.on(`${this.options.playerId}:${Events.PLAYER_RESIZE}`, this.playerResize, this)
```
leandromoreira: I think, maybe we should go first as a container event or core =/ but I dunno, thoughts @clappr/core ?

src/components/media_control/media_control.js
```diff
@@ -110,26 +125,36 @@ export default class MediaControl extends UIObject {
...
+    this.listenTo(this.container, Events.CONTAINER_ENDED, this.ended)
+    this.listenTo(this.container, Events.CONTAINER_VOLUME, this.onVolumeChanged)
+
+    var video = this.container.playback.el;
+    if (video.nodeName.toLowerCase() === 'video') {
+      // wait until the metadata has loaded and then check if fullscreen on video tag is supported
+      this.listenToOnce(this.container, Events.CONTAINER_LOADEDMETADATA, this.onLoadedMetadataOnVideoTag)
+
+      // watch for user to come out of fullscreen and remove the native controls that iOS tends to add
+      video.addEventListener('webkitendfullscreen', function() {
```
leandromoreira: I like this but we should handle this in a different manner, @towerz knows better.

kslimani: Maybe use `Events.CORE_FULLSCREEN` event boolean value ?

towerz: Indeed, @kslimani is right. There's already an event for fullscreen status change.

src/components/media_control/media_control.js
towerz: About the PLAYER_RESIZE event, perhaps we should change the `triggerResize` event on the core to trigger this event as a core event as well. This way we won't need to rely on the Mediator for events that are internal to the player instance.

src/components/media_control/media_control.js
bikegriffith: `CORE_FULLSCREEN` doesn't work on iOS when you're using the native controls to dismiss.  This is the only way I could find to let it work.

src/components/media_control/media_control.js
bikegriffith: FWIW, I didn't actually change any of this.  I just made the `if` block return above to remove a level of indentation.

src/components/media_control/media_control.js
kslimani: Good point. On iOS the event must be bind on the video element itself. I think (_but i did not test yet)_ that it should be possible to do late event binding somewhere in core if resolved playback container is a video element. @bikegriffith i will take a look this week to attempt to propose a fix for having CORE_FULLSCREEN trigger on iOS.

towerz: @bikegriffith There's a problem with this implementation. Perhaps it should be implemented on the HTML5Video playback itself. There's an option for the playback that tells it to use the default video tag controls (`playback.controls`), and this code would not follow the configuration correctly.

bikegriffith:
Shutting this down.  The implementation in this PR didn't really work on Android.  Not sure how to easily bend what we have in our fork to work here.

PR #1456, opened 2017-06-29
Improving playback tests

speres93:
Improving playback tests to reduce the complexity


leandromoreira:
OwW thank you so much @speres93 your help is much appreciated !!! 😄

PR #1460, opened 2017-07-12
Adds global register to Events

diegosperes:
Adds possibility to some plugin register a global event for anyone listen

### DIFF
--- a/src/base/events.js
+++ b/src/base/events.js
@@ -167,11 +167,11 @@ export default class Events {
   }
 
   static register(eventName) {
-    Events.custom || (Events.custom = {})
+    Events.Custom || (Events.Custom = {})
     let property = typeof eventName === 'string' && eventName.toUpperCase().trim()
 
-    if(property && !Events.custom[property]) {
-      Events.custom[property] = property.toLowerCase().split('_').map(
+    if(property && !Events.Custom[property]) {
+      Events.Custom[property] = property.toLowerCase().split('_').map(
         (value, index) => index == 0 ? value : value = (value[0].toUpperCase() + value.slice(1))
       ).join('')
     } else {
@@ -179,9 +179,9 @@ export default class Events {
     }
   }
 
-  static listCustomAvailable() {
-    Events.custom || (Events.custom = {})
-    return Object.keys(Events.custom).filter((property) => typeof Events.custom[property] === 'string')
+  static listAvailableCustomEvents() {
+    Events.Custom || (Events.Custom = {})
+    return Object.keys(Events.Custom).filter((property) => typeof Events.Custom[property] === 'string')
   }
 }
 
--- a/src/base/events.js
+++ b/src/base/events.js
@@ -165,8 +165,25 @@ export default class Events {
     }
     return this
   }
-}
 
+  static register(eventName) {
+    Events.custom || (Events.custom = {})
+    let property = typeof eventName === 'string' && eventName.toUpperCase().trim()
+
+    if(property && !Events.custom[property]) {
+      Events.custom[property] = property.toLowerCase().split('_').map(
+        (value, index) => index == 0 ? value : value = (value[0].toUpperCase() + value.slice(1))
+      ).join('')
+    } else {
+      Log.error('Events', 'Error when register event: ' + eventName)
+    }
+  }
+
+  static listCustomAvailable() {
+    Events.custom || (Events.custom = {})
+    return Object.keys(Events.custom).filter((property) => typeof Events.custom[property] === 'string')
+  }
+}
 
 /**
  * listen to an event indefinitely for a given `obj`
--- a/test/base/events_spec.js
+++ b/test/base/events_spec.js
@@ -1,9 +1,11 @@
 import Events from '../../src/base/events'
+import Log from '../../src/plugins/log'
 
 describe('Events', function(){
   beforeEach(function(){
     this.events = new Events()
     this.callback = sinon.spy()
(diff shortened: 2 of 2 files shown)


src/base/events.js
```diff
@@ -165,8 +165,22 @@ export default class Events {
...
+    let property = typeof eventName === 'string' && eventName.toUpperCase().trim()
+
+    if(property && !Events[property]) {
+      Events[property] = 'register:' + property.toLowerCase().replace(/_+/g, '.')
+    } else {
+      Log.error('Events', 'Error when register event: ' + eventName)
+    }
+  }
+
+  static listAvailables() {
```
towerz: This should be called `listAvailable`

src/base/events.js
```diff
@@ -165,8 +165,22 @@ export default class Events {
     }
     return this
   }
-}
 
+  static register(eventName) {
+    let property = typeof eventName === 'string' && eventName.toUpperCase().trim()
+
+    if(property && !Events[property]) {
+      Events[property] = 'register:' + property.toLowerCase().replace(/_+/g, '.')
```
towerz: I don't think custom event values should be prefixed with `register:`. Instead, we could use a camelized version of the event name.

For example:
WILL_CHANGE_SOURCE = 'willChangeSource'

You could do something like this:
```
"WILL_CHANGE_SOURCE".toLowerCase().split('_').map(function(value, index) { return index == 0 ? value : value = (value[0].toUpperCase() + value.slice(1)) }).join('')
```

Another suggestion is to put custom events under a Custom namespace, perhaps something like `Events.Custom.WILL_CHANGE_SOURCE`

test/base/events_spec.js
```diff
@@ -173,18 +126,77 @@ describe('Events', function(){
...
     this.events.on('clappr.any.event', handlerThatThrows)
-    this.events.on('clappr.any.event', handler)
+    this.events.on('clappr.any.event', this.callback)
     this.events.trigger('clappr.any.event')
 
-    expect(secondHandlerCalled).to.be.true
+    this.callback.should.have.been.calledOnce
+  })
+
+  it('when register, still call custom event', function(){
```
towerz: The test description should be the actual behavior the component should present but in present tense.

Something like `triggers custom event when registered`. The same works for the following tests.

test/base/events_spec.js
```diff
@@ -173,18 +126,77 @@ describe('Events', function(){
...
+  })
+
+  it('when register, set a namespace and put in default value', function(){
+    const eventName = 'PLUGIN_CUSTOM_EVENT'
+    Events.register(eventName)
+
+    expect(Events[eventName]).to.be.equal('register:plugin.custom.event')
+  })
+
+  it('when register, not override event when exist', function(){
```
towerz: `does not override an existing event`

test/base/events_spec.js
```diff
@@ -173,18 +126,77 @@ describe('Events', function(){
...
+    const eventName = 'PLAYBACK_READY'
+    Events.register(eventName)
+
+    expect(Events[eventName]).to.be.equal('playback:ready')
+  })
+
+  it('list all availables events without methods and default attributes', function(){
```
towerz: `list all available custom events`

test/base/events_spec.js
```diff
@@ -173,18 +126,77 @@ describe('Events', function(){
...
+    })
+
+    it('a empty string', function(){
+      Events.register('')
+      Events.register('      ')
+
+      this.stubLogError.should.have.been.calledTwice
+    })
+
+    it('different than string', function(){
```
towerz: `not a string`

test/base/events_spec.js
```diff
@@ -173,18 +126,77 @@ describe('Events', function(){
...
+    Events.register(eventName)
+    events = Events.listAvailables()
+
+    expect(events).to.include(eventName)
+    expect(events).to.include('PLAYBACK_READY')
+    expect(events).to.not.include('register')
+    expect(events).to.not.include('listEvents')
+  })
+
+  describe('not register when eventName is', function() {
```
towerz: `does not register an event when eventName is`

test/base/events_spec.js
```diff
@@ -173,18 +126,77 @@ describe('Events', function(){
...
+
+    beforeEach(function() {
+      this.stubLogError = sinon.stub(Log.getInstance(), 'error')
+    })
+
+    afterEach(function() {
+      Log.getInstance().error.restore()
+    })
+
+    it('a empty string', function(){
```
towerz: `an empty string`

src/base/events.js
```diff
@@ -165,8 +165,25 @@ export default class Events {
     }
     return this
   }
-}
 
+  static register(eventName) {
+    Events.custom || (Events.custom = {})
```
towerz: The custom events namespace should be `Custom`.

src/base/events.js
```diff
@@ -165,8 +165,25 @@ export default class Events {
...
+    if(property && !Events.custom[property]) {
+      Events.custom[property] = property.toLowerCase().split('_').map(
+        (value, index) => index == 0 ? value : value = (value[0].toUpperCase() + value.slice(1))
+      ).join('')
+    } else {
+      Log.error('Events', 'Error when register event: ' + eventName)
+    }
+  }
+
+  static listCustomAvailable() {
```
towerz: Perhaps this method could be called `listAvailableCustomEvents`.

PR #1486, opened 2017-08-09
Prevents wrong PLAYBACK_BUFFERFULL triggering

diegosperes:
Without this change, every time the `HTML5Video` playback is initialized and the variable `_bufferingState` is `undefined` it will trigger (incorrectly) the `PLAYBACK_BUFFERFULL` event.

### DIFF
--- a/src/playbacks/html5_video/html5_video.js
+++ b/src/playbacks/html5_video/html5_video.js
@@ -95,6 +95,7 @@ export default class HTML5Video extends Playback {
     super(...args)
     this._destroyed = false
     this._loadStarted = false
+    this._bufferingState = false
     this._playheadMoving = false
     this._playheadMovingTimer = null
     this._stopped = false
--- a/src/playbacks/html5_video/html5_video.js
+++ b/src/playbacks/html5_video/html5_video.js
@@ -88,14 +88,14 @@ export default class HTML5Video extends Playback {
    * @type Boolean
    */
   get buffering() {
-    return !!this._bufferingState
+    return this._isBuffering
   }
 
   constructor(...args) {
     super(...args)
     this._destroyed = false
     this._loadStarted = false
-    this._bufferingState = false
+    this._isBuffering = false
     this._playheadMoving = false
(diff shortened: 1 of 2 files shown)


test/playbacks/html5_video_spec.js
```diff
@@ -190,6 +190,14 @@ describe('HTML5Video playback', function() {
       expect(progress.current).to.be.equal(end[1])
     })
 
+    it('does not trigger PLAYBACK_BUFFERFULL when playback is initialized', function() {
```
leandromoreira:
```diff
+when the playback
-when playback
```

test/playbacks/html5_video_spec.js
```diff
@@ -190,6 +190,14 @@ describe('HTML5Video playback', function() {
...
+      let callback = sinon.spy()
+      this.playback.on(Events.PLAYBACK_BUFFERFULL, callback)
```
leandromoreira: Instead of using the created `playback` I'd create a new one here just to be sure that when it's initialized (aka instantiated, I guess) we're not facing this issue anymore.

```javascript
const playback = new HTML5Video(this.options)
```

leandromoreira:
Good catch! 👏 👏 👏 👏 👏 👏 👏

leandromoreira:
🍾 🍾 🍾 🍾 🍾 🍾 🍾 🍾

src/playbacks/html5_video/html5_video.js
```diff
@@ -95,6 +95,7 @@ export default class HTML5Video extends Playback {
     super(...args)
     this._destroyed = false
     this._loadStarted = false
+    this._bufferingState = false
```
towerz: I'd change this to `this._isBuffering`, or something like that. When talking about the `State` suffix, I understand it's something with specific values instead of a simple boolean flag.

diegosperes: I agree with you, I only maintained the old variable name, I changed this in latest commits :)

PR #1489, opened 2017-08-18
Add support to a seeked event

diegosperes:
Today Clappr does not support 'seeked' event, sometime we need listen this event to realize one action when seek is finished.

I talked with @towerz and we discuss about the new architecture of Clappr, so we decided to implement this event only for playback.

### DIFF
--- a/src/playbacks/html5_video/html5_video.js
+++ b/src/playbacks/html5_video/html5_video.js
@@ -64,7 +64,7 @@ export default class HTML5Video extends Playback {
       'playing': '_onPlaying',
       'progress': '_onProgress',
       'seeked': '_handleBufferingEvents',
-      'seeking': '_handleBufferingEvents',
+      'seeking': '_onSeeked',
       'stalled': '_handleBufferingEvents',
       'timeupdate': '_onTimeUpdate',
       'waiting': '_onWaiting'
@@ -337,6 +337,11 @@ export default class HTML5Video extends Playback {
     this._stopPlayheadMovingChecks()
     this._handleBufferingEvents()
     this.trigger(Events.PLAYBACK_PAUSE)
+  }
+
+  _onSeeked() {
+    this._handleBufferingEvents()
+    this.trigger(Events.PLAYBACK_SEEKED)
   }
 
   _onEnded() {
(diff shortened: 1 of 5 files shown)


src/playbacks/html5_video/html5_video.js
```diff
@@ -64,7 +64,7 @@ export default class HTML5Video extends Playback {
       'playing': '_onPlaying',
       'progress': '_onProgress',
       'seeked': '_handleBufferingEvents',
-      'seeking': '_handleBufferingEvents',
+      'seeking': '_onSeeked',
```
leandromoreira: `The seeking event is fired when a seek operation **began**.`

https://developer.mozilla.org/en-US/docs/Web/Events/seeking

Aren't we looking for the event after this?

We should tied this event to `seeked` https://developer.mozilla.org/en-US/docs/Web/Events/seeked

kslimani:
@diegosperes thank you for working on this feature! why only adds to playback ?

Currently, when playback **starts to seek**, the container event is [`CONTAINER_SEEK`](https://github.com/clappr/clappr/blob/master/src/components/container/container.js#L347).

The first issue, (imo) is that event is trigger by the container in seek() method : i think that container should listen to a playback `PLAYBACK_SEEK` event (_or PLAYBACK_SEEKING_) and bubble it up like other payback event. This is not a small modification, because you need adds the new event to all playbacks.

Then when playback has **finished to seek**, it should trigger a `PLAYBACK_SEEKED` event and container should listen to this event and then trigger a `CONTAINER_SEEKED` event. (_it require to modify all playbacks to add also this new event_).

PS: I am aware that this may be tricky as each playback event mechanisms differs and it **may** be only partially supported.

src/playbacks/html5_video/html5_video.js
diegosperes: @leandromoreira @vagnervjs Thanks guys, it's a mistake, I will fix it.

diegosperes:
Hi @kslimani, thank for your consideration, so in new Clappr architecture, @towerz was talked that the container not would be a proxy anymore, you can see this in android implementation, It's the reason of first implementation. But It's better change this only when Clappr will be updated to the new architecture, so I implemented the CONTAINER_SEEKED too.

I think is better do this way ` it may be only partially supported`

src/playbacks/html5_video/html5_video.js
diegosperes: Done! :)

kslimani:
Allright, so with these changes, the seeked event is triggered by playback and bubble up by container and the seek(ing) event is still triggered only by container. (_in seek method_).

I will take a look at android implementation, thank you for pointing this.

PR #1500, opened 2017-08-29
fix(build): add templated public path to Webpack url-loader module config

### LINKED ISSUE #1480: FlasHLS locate .swf to local domain
Hello sir,

FlasHLS doesn't locate .swf to baseUrl.
Here my code : 
```
	  source: 'http://dev.example.me/hls/stream.m3u8',
	  baseUrl: 'dist',
	  mute: true,
	  height: 360,
	  width: 640,
	  plugins: [Clappr.FlasHLS]
	});
```
When i view console network tab, it's request url to http://localhost/[number].swf
Not to http://localhost/dist/[number].swf

Thank you before

kslimani:
Should fix #1480.

Previously it was working because the `url-loader` module internally use the `file-loader` module where public path default to webpack `__webpack_public_path__` global variable (_which is Webpack `output.publicPath` value_).

It seems that if you set the `publicPath` configuration parameter to `url-loader`, it is also passed to `file-loader`. So we can pass here the templated value and it is resolved at runtime instead of compilation time if set in `output.publicPath` (_i think it is because Clappr is writted in ES6_).

### DIFF
--- a/src/playbacks/base_flash_playback/base_flash_playback.js
+++ b/src/playbacks/base_flash_playback/base_flash_playback.js
@@ -40,6 +40,7 @@ export default class BaseFlashPlayback extends Playback {
   }
 
   render() {
+    this.$el.attr('data', this.swfPath)
     this.$el.html(this.template({
       cid: this.cid,
       swfPath: this.swfPath,
--- a/test/playbacks/flash_spec.js
+++ b/test/playbacks/flash_spec.js
@@ -40,4 +40,11 @@ describe('Flash playback', function() {
     this.callback.should.have.been.calledOnce
     expect(this.flash.isReady).to.be.true
   })
+
+  it('should update data element attribute with base url on render', function() {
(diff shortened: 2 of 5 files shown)


asskovalev:
this fix did not help
`object` element still renders with `data="/809981e5b09d5336c45d72d0869ada2a.swf"` attribute
looks like FlasHLS plugin renders before baseUrl is initialized

kslimani:
@asskovalev you are correct, there is an additionnal issue i missed with Internet Explorer. (_i tested only Firefox and Chrome_). I will perform additionnal tests with Windows and report back hopefully with a fix.

kslimani:
@asskovalev i added an additional fix, could you please check again ?
@towerz could you also check if it does not introduce breaking changes (_especially with the changes you made to fix Firefox crashs_), i will also perform tests.

asskovalev:
looks great. thanks!

towerz:
This seems to be working for me as well.

PR #1504, opened 2017-09-02
Update jsDelivr CDN links

MartinKolarik:
First, congrats - [clappr receives more than 400 M requests per month at jsDelivr](https://www.jsdelivr.com/package/npm/clappr), which makes it the 11th most popular package we host!

While our new back-end supports serving files from both npm and GitHub, we recommend always using only one of these methods for every package, with npm being the preferred one (using only npm means you can get detailed package usage statistics and npm packages are also searchable on our website).

I updated the links to use our npm endpoint.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@
 Add the following script on your HTML:
 ```html
 <head>
-  <script type="text/javascript" src="https://cdn.jsdelivr.net/gh/clappr/clappr@latest/dist/clappr.min.js"></script>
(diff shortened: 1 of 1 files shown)


kslimani:
@leandromoreira isn't the index public page supposed to load the current clappr build while running webpack dev server ?

leandromoreira:
@kslimani yeah, I was thinking about the demo page which could benefit from it too.

kslimani:
@leandromoreira : indeed, Clappr demo page it should (imo) use npm jsDeliver url 👍

MartinKolarik:
Hey @leandromoreira, can we get this merged?

flavioribeiro:
done!

PR #1511, opened 2017-09-26
displaying seek time and duration according #EXT-X-PROGRAM-DATE-TIME

otmjka:
Hi there, 

The attempt to befriend #EXT-X-PROGRAM-DATE-TIME and actualLiveTime option.

### DIFF
--- a/src/plugins/seek_time/seek_time.js
+++ b/src/plugins/seek_time/seek_time.js
@@ -26,19 +26,20 @@ export default class SeekTime extends UICorePlugin {
   get mediaControl() { return this.core.mediaControl }
   get mediaControlContainer() { return this.mediaControl.container }
   get isLiveStreamWithDvr() { return this.mediaControlContainer && this.mediaControlContainer.getPlaybackType() === Playback.LIVE && this.mediaControlContainer.isDvrEnabled() }
-  get durationShown() { return this.isLiveStreamWithDvr && !this.useActualLiveTime }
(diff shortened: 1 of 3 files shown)


src/plugins/seek_time/seek_time.js
```diff
@@ -82,10 +84,19 @@ export default class SeekTime extends UICorePlugin {
...
+      if (this.actualLiveServerTimeDiff) {
+        d = new Date(new Date().getTime() - this.actualLiveServerTimeDiff)
+        e = new Date(d)
+        secondsSinceMidnight = (e - d.setHours(0,0,0,0)) / 1000
+      } else if (this.firstFragDateTime) {
+        e = new Date(this.firstFragDateTime)
+        d = new Date(this.firstFragDateTime)
+        timezoneOffset = e.getTimezoneOffset() * 60
+        d.setHours(0,0,0,0)
+        secondsSinceMidnight = ((e.getTime() - d.getTime()) / 1000) + timezoneOffset + this.duration
```
towerz: @otmjka there seems to be a small problem here. I believe the timezoneOffset shouldn't be added to the actual time, since the local timezone is already considered when creating the date objects, and setting the time to midnight.

When you add this, you actually offsets the time to be displayed as UTC. At least it's what is happening to me, since I'm at UTC-2 and I'm seeing times advanced by 2 hours.

PR #1521, opened 2017-10-17
Enable tree shaking - closes #1383

### LINKED ISSUE #1383: export using the `modules` attribute
Here is an example of what this looks like on another project: https://github.com/uber/deck.gl/blob/master/package.json#L16-L17

By adding this attribute in the `package.json` you enable modern browsers/node/webpack to do code splitting, which drastically reduces the size of the output by only including exactly what is being used.

Here is what `clappr` looks like right now (code splitting disabled, because the `main` in the package.json is the pre-compiled file):
screen shot 2017-05-02 at 11 26 59 am

For somebody who is just using the player and supporting a few video types, this is a pretty huge amount of unused code.

I think for this project it would be as simple as adding `module: "src/index.js"` to enable webpack/rollup/whatever users to code split - happy to send a PR.

contra:
Would appreciate a publish so this can be used

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "version": "0.2.75",
   "description": "An extensible media player for the web",
   "main": "./dist/clappr.js",
-  "module": "./src/index.js",
+  "module": "./src/main.js",
   "scripts": {
     "release": "webpack --progress --output-filename clappr.min.js",
     "build": "webpack --progress",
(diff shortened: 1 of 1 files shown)


contra:
Should ignore non-js files from coverage reports ^

contra:
After more testing - this will require a rethinking of the way the build works currently. The aliases, svgs, and CSS loading is going to make this a breaking change that would require all webpack users to write those rules themselves. It doesn't look like there is any way to tell webpack that it should use the config of a dependency to manage that dependencies assets.

leandromoreira:
Thanks, @contra I think @thiagopnts is working on this too!

PR #1525, opened 2017-10-30
Fix HLS playback options

vagnervjs:
Get `hlsjsConfig` from `this.options.playback.hlsjsConfig` instead `this.options.playback`.


src/playbacks/hls/hls.js
```diff
@@ -449,7 +449,7 @@ export default class HLS extends HTML5VideoPlayback {
     // seeks to areas after this point sometimes have issues
     if (this._playbackType === Playback.LIVE) {
       let fragmentTargetDuration = data.details.targetduration
-      let hlsjsConfig = this.options.playback || {}
+      let hlsjsConfig = this.options.playback.hlsjsConfig || {}
```
shaharmor: Why make this change?

PR #1527, opened 2017-11-01
Update External_Plugins.md

Lethea:
Clapper js scroll text, marquee overlay plugin added.


leandromoreira:
Cool, thank you @Lethea but where are the dependencies? `package.json` ( I think you relied on jQuery isn't?) where is the build/minified version?

Lethea:
Hi @leandromoreira , there is no need to build anything, I developed this by following first plugin guide. Therefore there is no build etc script generated . If needed, I will download the plugin generator and develop this with this generator by the way the skeloton will be appear as you ask ( dependicies build script src etc )  :) However fordeveloper  this sample help them. if you accept this , the next time I'll develop it with plugin-generator.see you

leandromoreira:
Thank you very much !!!

PR #1528, opened 2017-11-01
fix(core): Append core element to parent only once

vagnervjs:
The core element needs to be immutable in DOM to avoid render problems and others unexpected behaviors (e.g. DFP freezed after a new load on Android)

### DIFF
--- a/src/components/core/core.js
+++ b/src/components/core/core.js
@@ -245,7 +245,7 @@ export default class Core extends UIObject {
     this.renderContainers()
     this.setupMediaControl(this.getCurrentContainer())
     this.render()
-    this.$el.appendTo(this.options.parentElement)
+    this.appendToParent()
     return this.containers
   }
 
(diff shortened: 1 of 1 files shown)


vagnervjs:
[[DFP] Rever implementação para evitar AD "congelado" no Android com autoplay](https://trello.com/c/0F2120Ax/469-dfp-rever-implementa%C3%A7%C3%A3o-para-evitar-ad-congelado-no-android-com-autoplay)

src/components/core/core.js
```diff
@@ -349,6 +349,11 @@ export default class Core extends UIObject {
     }
   }
 
+  appendToParent() {
+    let hasCoreParent = this.$el.parent() && this.$el.parent().length
```
leandromoreira: Is there any possibility of `$el` or `parent()` return `undefined`?

towerz: `$el` is valid just after the `UIObject` constructor is called, so it would be `undefined` only if some rogue component destroys the element. The `parent()` call can return undefined if the core element is removed from DOM or `destroy`ed.

towerz: I think it's safe to assume that $el is always valid.

PR #1532, opened 2017-11-07
Manage _ready() state on HLS playback

joaopaulovieira:
Playback ready state treatment for not use html5 playback method.

Also, this pull request normalize the name of the setup methods between playbacks. The [dash playback](https://github.com/clappr/dash-shaka-playback) plugin already follows [this naming](https://github.com/clappr/dash-shaka-playback/blob/master/src/clappr-dash-shaka-playback.js#L257).

### DIFF
--- a/src/playbacks/hls/hls.js
+++ b/src/playbacks/hls/hls.js
@@ -15,6 +15,7 @@ export default class HLS extends HTML5VideoPlayback {
   get name() { return 'hls' }
 
   get levels() { return this._levels || [] }
+
   get currentLevel() {
     if (this._currentLevel === null || this._currentLevel === undefined) {
       return AUTO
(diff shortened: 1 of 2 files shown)


leandromoreira:
Thanks, @joaopaulovieira can you also fix the tests?

```
Firefox 56.0.0 (Linux 0.0.0) HLS playback hls.js configuration should use hlsjsConfig from player options as fallback FAILED
	playback._setupHls is not a function
	@test/playbacks/hls_spec.js:21030:7
```

It seems that you only need to call `_setup` instead of `_setupHls` on the test. I think maybe worth to do a quick `git grep _setupHls` to see if this is used in anywhere else.

joaopaulovieira:
Yes! What a failure!
I will correct it now :see_no_evil:

joaopaulovieira:
[Seletor de áudio para playback HLS](https://trello.com/c/Z8BUAWGx/552-seletor-de-%C3%A1udio-para-playback-hls)

PR #1539, opened 2017-12-04
Use Autoprefixer instead node-bourbon mixins

joaopaulovieira:
At the time of build for the project start or to run the test stack, warnings are appearing for me about mixins that Bourbon will deprecated in its new version `5.x.x`. 

In the repository of wrapper for Bourbon `node-bourbon` [has a issue about it](https://github.com/lacroixdesign/node-bourbon/issues/37) and it seems that the project is not using the most current version of Bourbon.

In the Bourbon repository there is an [issue talking about the project option](https://github.com/thoughtbot/bourbon/issues/702) and, according to them, `... we've decided to remove browser prefixing features from Bourbon.`

Alternatively, they suggest post-processors and the [Autoprefixer](https://github.com/postcss/autoprefixer) are best for this.

With this, I configured [postcss-loader](https://github.com/postcss/postcss-loader) in the project and removed the node-bourbon dependency.

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,7 +64,6 @@
     "lolex": "^2.0.0",
     "mkdirp": "^0.5.1",
     "mocha": "^3.3.0",
-    "node-bourbon": "^4.2.8",
     "node-sass": "^4.5.3",
     "postcss-loader": "^2.0.9",
     "s3": "^4.1.1",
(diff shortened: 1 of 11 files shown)


towerz:
🍻

Issue #1325, opened 2017-03-13
Zepto collection assertion bug in UiObject


kslimani:
I think there is maybe a _minor_ error in [setElement](https://github.com/clappr/clappr/blob/master/src/base/ui_object.js#L145) method in `UIObject` : i suspect that `element instanceof $` never return true.

According [this comment](https://github.com/madrobby/zepto/issues/349#issuecomment-4985091), it should be something like :

```javascript
this.$el = $.zepto.isZ(element) ? element : $(element)
```

But i am not sure about this. Can someone from Clappr core confirm ?

kslimani:
Closing issue as it is fixed by #1541.

PR #1541, opened 2017-12-05
Fix is element instance of zepto

stanogurnik:
Small fix: `instanceof $` does not work with zepto as it does with jQuery, zepto has its own check `$.zepto.isZ(e)` -> https://github.com/madrobby/zepto/issues/349


kslimani:
yup, i had the same conclusion : ref #1325

towerz:
Indeed, you're right @stanogurnik. Thanks!

PR #1542, opened 2017-12-05
Fix documentation of `mute` option + Ensure setting native `muted` property in Playback plugin(s)

tchakabam:
There was a typo here. It's `muted` not `mute`.

Now this option actually will work I guess :)

Also, I hope that we can be able to allow setting the HTML5 muted property safely before any playback to allow muted autoplay in Safari 11 (see https://bitmovin.com/play-not-play-new-autoplay-policies-safari-11-chrome-64/)

I am aware that in fact Clappr shares concept of "muted" and "volume === 0". which i think is a good thing to do in order to keep things actually clearer and simpler for applications.

Here in the case of HTML5 playback, there is this native property, and it is important to set it as configured before first play() in order to allow auto-play on platforms that only allow that when it is muted (not necessarily native volume property zero but actually muted and defaultMuted set).

### DIFF
--- a/src/playbacks/html5_video/html5_video.js
+++ b/src/playbacks/html5_video/html5_video.js
@@ -117,6 +117,8 @@ export default class HTML5Video extends Playback {
     }
 
     $.extend(this.el, {
+      muted: this.options.muted,
+      defaultMuted: this.options.muted,
       loop: this.options.loop,
       poster: posterUrl,
       preload: preload || 'metadata',
--- a/src/playbacks/html5_video/html5_video.js
+++ b/src/playbacks/html5_video/html5_video.js
@@ -252,10 +252,18 @@ export default class HTML5Video extends Playback {
     this.el.volume = value / 100
   }
 
+  /**
+   * @deprecated
+   * @private
+   */
   mute() {
     this.el.muted = true;
   }
 
+  /**
+   * @deprecated
+   * @private
+   */
   unmute() {
     this.el.muted = false;
   }
--- a/src/playbacks/html5_video/html5_video.js
+++ b/src/playbacks/html5_video/html5_video.js
@@ -252,6 +252,18 @@ export default class HTML5Video extends Playback {
     this.el.volume = value / 100
   }
 
+  mute() {
+    this.el.muted = true;
+  }
+
+  unmute() {
+    this.el.muted = false;
+  }
+
+  isMuted() {
+    return !!this.el.volume
+  }
+
   isPlaying() {
     return !this.el.paused && !this.el.ended
   }
--- a/src/playbacks/html5_video/html5_video.js
+++ b/src/playbacks/html5_video/html5_video.js
@@ -252,18 +252,6 @@ export default class HTML5Video extends Playback {
     this.el.volume = value / 100
   }
 
-  mute() {
-    this.el.muted = true;
-  }
-
-  unmute() {
-    this.el.muted = false;
-  }
-
-  isMuted() {
-    return !!this.el.volume
-  }
-
   isPlaying() {
     return !this.el.paused && !this.el.ended
   }
--- a/src/playbacks/html5_video/html5_video.js
+++ b/src/playbacks/html5_video/html5_video.js
@@ -253,11 +253,11 @@ export default class HTML5Video extends Playback {
   }
 
   mute() {
-    this.el.volume = 0
+    this.el.muted = true;
   }
 
   unmute() {
-    this.el.volume = 1
+    this.el.muted = false;
   }
 
   isMuted() {
--- a/src/playbacks/html5_video/html5_video.js
+++ b/src/playbacks/html5_video/html5_video.js
@@ -257,7 +257,7 @@ export default class HTML5Video extends Playback {
    * @private
    */
   mute() {
-    this.el.muted = true;
+    this.el.muted = true
   }
 
   /**
@@ -265,7 +265,7 @@ export default class HTML5Video extends Playback {
    * @private
    */
   unmute() {
-    this.el.muted = false;
+    this.el.muted = false
   }
 
   isMuted() {
--- a/src/playbacks/html5_video/html5_video.js
+++ b/src/playbacks/html5_video/html5_video.js
@@ -117,8 +117,8 @@ export default class HTML5Video extends Playback {
     }
 
     $.extend(this.el, {
-      muted: this.options.muted,
-      defaultMuted: this.options.muted,
+      muted: this.options.mute,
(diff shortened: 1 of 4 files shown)


src/playbacks/html5_video/html5_video.js
```diff
@@ -251,11 +253,11 @@ export default class HTML5Video extends Playback {
   }
 
   mute() {
```
tchakabam: it's unclear to me: are these methods even used somewhere? :)

they aren't part of the Playback interface definition, correct?

i am aware that in fact Clappr shares concept of "muted" and "volume === 0". which i think is a good thing to do in order to keep things actually clearer and simpler for applications.

here in the case of HTML5 playback, there is this native property, and it is important to set it as configured before first play() in order to allow auto-play on platforms that only allow that when it is muted (not necessarily native volume property zero but actually muted and defaultMuted set).

leandromoreira:
We use [`mute`](https://github.com/clappr/clappr/blob/master/doc/BUILTIN_PLUGINS.md#mute) as part of our API, maybe it's better just to fix this on the rendering or offer both `muted` and `mute` for a while.

This change will break an API.

leandromoreira:
@clappr/core ^

tchakabam:
sorry, which API?

I can only see `muted` in other parts of Clappr code :) 

For sure we can add `mute` in the Clappr options as equivalent property and rewrite it to `muted` internally? :) 

All for making this as easy as possible for anyone, but it should be agreed I guess.

tchakabam:
@leandromoreira Documentation was my ground truth: https://github.com/clappr/clappr/blob/master/src/components/player.js#L135

towerz:
@tchakabam indeed, there's a typo in the documentation. The `mute` option has always been the right one. This should get fixed as well.

tchakabam:
@towerz OK. Wether the mistake is in the documentation or in the code was the question I guess :)

We assumed the documentation to be correct! Therefore I had to assume the code is wrong.

Sorry for the misunderstanding.

We are usually assuming that the documentation is actually correct. So I will probably update the docs :)

However, what we still will need is setting `muted` and `defaultMuted` on the video element right from the start.

kslimani:
@tchakabam if you need player to be muted right from the start, it should be possible. There is an example [here](https://github.com/clappr/clappr/pull/1432).

tchakabam:
@kslimani Cool, thanks! :) 

However, this is at least fixing the wrong documentation which was stating that the Clappr.Player option is `muted`, while media-control was assuming it is `mute`. Since everyone seems to agree on `mute` and it was the only thing working, the documentation has now been updated (whoever tried to use `muted` before didn't get a result anyway, so this isn't like it's a breaking spec change or so :) )

Also, I think it would be nicer to also for HTML5 video make sure that the video element `muted` and `defaultMuted` properties are set on creation of Playback. Because otherwise we are relying on side effect from MediaControl to take care of setting volume to zero in case of muted. It is better if this interaction is not necessary and the Playback already can set up the respective native properties itself, this is how I would expect it to interpret the Player options like. What do you think @towerz @leandromoreira @kslimani @tjenkinson ? :)

tchakabam:
@tjenkinson I agree with that, there only needs to be one place where we set `muted` based on volume, and everything else can drive that. Will change that.

But what are the `mute` and `unmute` methods really used for, i can not see them being called anywhere. Or is it just through `core.getCurrentPlayback()...` ? It is not part of the `Playback` interface, so what do they do here really? Some things aren't quite clear here I'd say :)

We would however really like of the HTML5 plugin here would be more self-contained concerning how it applies the native video element options, meaning that it should not have to rely on MediaControl to interpret the player options and set the volume to zero "sometimes". If the play for some reason happens before the volume call, we are screwed, this holds better ;) Or if someone disable MediaControl plugin or implements it's own. Playback's should be self-contained what this concerns I would argue.

towerz:
@tchakabam you're right. Actually, the playback should be responsible for these kind of operation and semantic.

About the `mute`/`unmute` operations, this should be called from the public `Player` interface. There are `mute`/`unmute` operations there, but as you pointed out, they're not calling their respective playback methods, but instead, they're changing volume to/from 0 using the MediaControl.

About the MediaControl being disabled/overridden, you're also correct. It shouldn't be responsible for those operations, only for translating the user interface interactions to playback operations. This will be changed soon, since the MediaControl will no longer be a special component, but instead it will be a `UICorePlugin`.

src/playbacks/html5_video/html5_video.js
```diff
@@ -250,18 +252,6 @@ export default class HTML5Video extends Playback {
     this.el.volume = value / 100
   }
 
-  mute() {
```
towerz: I'd rather keep this for now, since there could be some plugin implementations that rely on this. 

Perhaps instead of removing these methods, we could fix them to better reflect the `muted`/`defaultMuted` properties of the element.

tchakabam: yes, that's what I thought. @tjenkinson they could call just the plugins `volume` method but then we d have to store a state of muted volume to recover it on unmute. I'd rather avoid that at this point.

maybe just have them set `this.el.muted` accordingly.

and also we could set them `@deprecated` and `@private` as docs decorator to show that this is not part of `Playback` API and that they may be removed in a minor version bump.

towerz: I agree, we could just have those methods set `this.el.muted`, and make those methods `@deprecated` and `@private`.

We're still deciding whether or not to change the `Playback` API to include the `muted` state, but with a more concise interface.

tchakabam:
@towerz ok, nice!

> We're still deciding whether or not to change the Playback API to include the muted state, but with a more concise interface.

My thought here is that it is actually a good thing to have one single method, `volume`, that one needs to implement to have a valid `Playback` implementation, and that `0` also drives eventual native muting functionnality. I couldn't really think about a case where I would need to differentiate there, and also it motivates to keep muted volume caching and recovery outside of the plugin and at a higher level, for example in like it is now MediaControl.

At least for this PR, I'd definitely prefer to just leave at the docs fix and the initial native `muted` and `defaultMuted` handling to have the spec aligned and functionnality safe.

I am happy to help with improving the interface if we find an agreement but as I said I think it's actually very good like it is, given that it is implemented like here.

towerz:
Thanks, guys! 🍻

rromanchuk:
Couple years late to the party, but any idea instantiation with `mute: true,`  followed by `player.unmute()`, followed by `player.configure(src: "new hls stream")` would re-apply the muted attribute? Do I need to be tracking the state myself just to load a new stream?

Issue #839, opened 2016-02-16
How to create custom Media Control?


mikeevstropov:
If you have some example, please show it.
I tryed:
```javascript
var element = document.getElementById('player');
var MyMediaControl = function () {
	this.template = function () { return 'hello'; }
	this.stylesheet = function () { return '#player { background: blue }'; }
}
var player 	= new Clappr.Player({
	'source' 		: src,
	'width'  		: 480,
	'height' 		: 240,
	'mediacontrol'	: MyMediaControl
});
player.attachTo(element);
```
or
```javascript
Object.defineProperty(Clappr.MediaControl, "template", {
	get: function() { return 'Hello!' }
});		  
Object.defineProperty(Clappr.MediaControl, "stylesheet", {
	get: function() { return 'body { background: blue }' }
});
```
or
```javascript
Clappr.MediaControl.prototype.template.source = 'hello';
Clappr.MediaControl.prototype.stylesheet.source = 'body { background: blue }';
```
But it is not working.
Thank you!

https://kangax.github.io/compat-table/es6/

leandromoreira:
Hi @mikeevstropov I think you can use [this example to guide](https://github.com/clappr/clappr/wiki/Our-first-plugin) you.

mikeevstropov:
@leandromoreira Thank you for answer. But i need edit player ui like template. It's tutorial real right way?

leandromoreira:
@mikeevstropov you can also (or at least should be able to):
```javascript
var HelloWorld = Clappr.MediaControl.extend({})
// https://github.com/clappr/clappr/#control-bar-colors
  var player = new Clappr.Player({
    source: "http://your.video/here.mp4",
    mediacontrol: HelloWorld
  });
```

@clappr/core do you have any example of custom media control?

mikeevstropov:
@leandromoreira Sorry for my bad eng. Last example look like what i want. But `extend` is not defined in `MediaControl`. I use video.js and videojs-contrib-hls plugin. But m3u8 stream with videojs is very bad solution for production (amazing huge busg). I try set my vjs payer template into Clappr.

mikeevstropov:
Wow, it is not working too...
```javascript
var element = document.getElementById( 'player' );

class MyMediaControl extends Clappr.MediaControl {
	get template() {
		return Clappr.template(
			`
			<div class="media-control-background" data-background></div>
				<div class="media-control-layer" data-controls>
					Where is this string??? Wtf?
				</div>
			</div>
			`
[...]

PR #1544, opened 2017-12-06
Update BUILTIN_PLUGINS.md

bent0b0x:
According to [this issue](https://github.com/clappr/clappr/issues/839#issuecomment-206095423) the current documented way to specify custom media controls is incorrect.


leandromoreira:
Actually, it should be simpler yet like `plugins: [AnyPlugin]`, isn't @towerz ?

towerz:
Thanks @bent0b0x, nice catch!

@leandromoreira you're right. I think we'll have this soon, since I'm back to working on this project, at least part-time.

PR #1545, opened 2017-12-07
"Lighther" build only with plain html5 plugins

tchakabam:
In our product we are using a fork where we do something like this in order to get a lighter build :)

We don't always need Hlsjs, and we don't need all the Flash stuff.

We are lowering the Clappr non-minified build from > 1Mb to ~400 Kbytes.

Minified goes down to from ~500 to ~200Kbytes.

We already load a lot of stuff on the frontend, and people are really unhappy about such a big library full of stuff that we don't necessarilly use.

So it would be great if such a build option could be integrated into our build system.

It's a proposal on how to do it, all open for discussion, as long as we hit the goal of being able to produce easily a build without all "non-essential" playback plugins :)

BTW my personal opinion is that all these plugins:

1) Yes can be shipped in the main Clappr git repo, it has a lot of advantages of having the core and the plugins in one repo (for example CI / testing)

2) but should be shipped as individually built modules, that one can include or not into the page, or load on demand


leandromoreira:
Cool @tchakabam we discussed that a long time ago and now we're doing it again, @towerz was looking into how to make it modular (breaking the playbacks into their own github plugin and changing how we build, with two versions full and light) and @thiagopnts was working individually to shrink clappr size too.

tchakabam:
as i said, i think it's good to have one repository, instead of having to deal with different repos.

so i would strongly favor for keeping the one big repo, but allowing the build system for one core, and many plugin binaries. in that way, you can distribute one version of everything that will actually work together, instead of always having to match the versions here and there, that's hell for anyone who wants to use it. with the mono-repo, if core breaks a plugin API, we see in CI tests on the one repo :)

It's just that you need to set up the build system to spit out several binaries :)

that's pretty easy with webpack in fact, with the chunks mode.

also, what we did in the playback middleware in my current project is that we load dependencies on-demand, using `loadjs` and dynamic `import`.

because sometimes we use Shaka and sometimes a proprietary SDK on top of Clappr, and depending on the platform we load either of the two dependent libs (and eventually the ShakaPlayback plugin).

one could make the `Loader` here an actual loader by allowing to get the plugins on-demand, and resolving promises to their constructors once they are loaded.

Btw there are a few other projects out there who do it so, for example GStreamer: https://cgit.freedesktop.org/gstreamer/

And Google also has a mono-repo, even if we re not there yet :speak_no_evil:

tchakabam:
But anyway, I guess the broader discussion is one thing, this is just a small step in order to allow people having a lighter build, and we could distribute it as a seperate webpack output bundle (different than code-splitting/chunks), just by running these two configs and having a file `clappr.plainhtml5only[.min].js`

src/components/loader/loader.js
```diff
@@ -48,7 +48,21 @@ export default class Loader extends BaseObject {
...
     this.playerId = playerId
-    this.playbackPlugins = [HLSVideoPlayback, HTML5VideoPlayback, HTML5AudioPlayback, FlashVideoPlayback, FlasHLSVideoPlayback, HTMLImgPlayback, NoOp]
+    this.playbackPlugins = [
+      HTML5VideoPlayback,
+      HTML5AudioPlayback,
+      HTMLImgPlayback,
+      NoOp
+    ]
+
+    if (!PLAIN_HTML5_ONLY === 'yes') {
```
towerz: Why don't we use true/false here? This way the `if` statement would be much simpler:
```
if (PLAIN_HTML5_ONLY)
```

webpack.config.js
```diff
@@ -22,6 +22,19 @@ if (process.env.npm_lifecycle_event === 'release') {
   webpackConfig.plugins.push(new Clean(['dist'], {verbose: false}))
 }
 
+webpackConfig.devtool = 'source-map';
+
+const voidModulePath = path.resolve('./src/base/void');
+
+if (process.env.CLAPPR_PLAIN_HTML5_ONLY === 'yes') {
```
towerz: Same thing here, we could use `true`/`false`.

src/components/loader/loader.js
tchakabam: yes, but it's tricky i think... :) we are getting this from the environment, so per se it is a always going to be a `string`. Shell rules are that we have to give it some value: env var can't be empty i think, that's same as not existing, so it will simply not exist in `process.env`. And when we do Boolean(someString) it will be `true` except if the string is zero length. What we could do in the webpack file would be: `Boolean(process.env.CLAPPR_...)` which will result to `true` for any value of the env var. Or we do `(process.env.CLAPPR_... === 'yes')` or `(process.env.CLAPPR_... === 'true')`, which will force users to certain values. I think that's better. What do you prefer? 

I guess the deeper question really is how to deal with env vars in your build conf.

towerz: I think we can hold off a little bit the generic solution and reach a middle ground for now. 

Currently, we're trying to have a PLAIN_HTML5_ONLY build and the default one, with all plugins. So, we just need to enable the PLAIN_HTML5_ONLY flag. In the `Loader` code, I believe it's better to have it as a boolean value since we don't need to have the possible string values of the environment variable there.

As for the `process.env.CLAPPR_...` variable, we could either use true/false for it, having the empty/non-existent/false values to disable it and using whatever else for true, or we can force it to have a yes/no value. I think the first option would be better, since the DefinePlugin allows boolean values, AFAIK.

This way, instead of having:
```
new webpack.DefinePlugin({
  VERSION: JSON.stringify(require('./package.json').version)
  PLAIN_HTML5_ONLY: JSON.stringify(process.env.CLAPPR_PLAIN_HTML5_ONLY)
})
```
we could have:
```
new webpack.DefinePlugin({
  VERSION: JSON.stringify(require('./package.json').version)
  PLAIN_HTML5_ONLY: !!process.env.CLAPPR_PLAIN_HTML5_ONLY
})
```

towerz: Just to make things easier, we could also add scripts that set this environment var automatically, like:
```
"scripts": {
    "build": "webpack --config webpack.dev.js --progress",
    "build:plainhtml5": "CLAPPR_PLAIN_HTML5_ONLY=true webpack --config webpack.dev.js --progress",
   ...
}
```

tchakabam: OK, totally agreeing! :+1:

tchakabam:
@towerz OK, this should be what we want? :)

Also, should we commit the source-map for release builds? I guess it makes sense to distribute it along with the package (since we are distro'ing `src` along with the package), so that one can actually debug even when using minified build.

And for the future roadmap, it would be good to:

* Have a more generic solution for build-option flags via env vars as discussed above
* Have only one single call to webpack, and differentiate release/debug/other cases via the build options flags

towerz:
@tchakabam it makes sense to add source maps to releases. I thought we were generating it, but we're not.

tchakabam:
Source-maps were enabled only in the UglifyJS config, but it also needs to be set as `devtool` in main webpack config.

tchakabam:
@towerz Source-maps were generated but only when doing uglification, and by uglify-js. The `devtool` option on the main webpack config makes sure that generally you get a mapping from webpack bundle to original module sources, no matter if it has been uglified or not.

so there is a `.min.map` that maps from the uglified code, and one that maps from the unminified code, both towards the original source modules now.

tchakabam:
@towerz After adding the source-maps for non-min build, I need to increase test coverage, so I fixed/added some tests in HLS and media-control specs.

tchakabam:
@towerz Sure, that makes sense! I removed all of it now. But what about the source-map for non-minified builds? :)

tchakabam:
@towerz I just thought: Wouldn't it be good to also distro the smaller build right away in the repo? for example with a naming patterns like `clappr.plainhtml5only{.min}.js` ?

towerz:
@tchakabam that's a good idea, can you add this?

tchakabam:
@towerz it's done :beers: merry xmas!

towerz:
Sorry for taking so long. Thanks @tchakabam!

PR #1546, opened 2017-12-07
Eslint fix option + rm jscs 

tchakabam:
* Add ESlint `--fix` script 
* Remove JSCS scripts (project has merged with ESlint, see http://jscs.info/)
* Remove old jshint dependency


tchakabam:
i guess we should maybe just make sure that we are covering the jscs rule with the eslint ones already

tchakabam:
@leandromoreira I guess although I didn't write any functionnality, I am gonna have to write some test now :D

.eslintrc.js
```diff
@@ -44,6 +44,12 @@ module.exports = {
             "error",
             "never"
         ],
-        "no-var": "error"
+        "no-var": "error",
+        // part of Node Style-guide but ignored
+        "max-len": 0,
+        "max-statements": 0,
+        "object-curly-spacing": 0,
+        "brace-style": 0,
```
towerz: We use the "1tbs" brace style. 

I recommend changing this to  ["error", "1tbs", { "allowSingleLine": true }]

.eslintrc.js
```diff
@@ -44,6 +44,12 @@ module.exports = {
             "error",
             "never"
         ],
-        "no-var": "error"
+        "no-var": "error",
+        // part of Node Style-guide but ignored
+        "max-len": 0,
+        "max-statements": 0,
+        "object-curly-spacing": 0,
```
towerz: Let's change this to ["error", "always"]. We usually add padding to single line object literals.

src/components/core/core.js
```diff
@@ -322,9 +322,9 @@ export default class Core extends UIObject {
     this.getCurrentContainer().trigger(showing?Events.CONTAINER_MEDIACONTROL_SHOW:Events.CONTAINER_MEDIACONTROL_HIDE)
 
     if (showing)
-      this.$el.removeClass('nocursor')
+    {this.$el.removeClass('nocursor')}
```
towerz: One of the things we want to avoid is doing this. We should either have the braces in "1tbs" style or have a single-line if statement, preferably with no braces.

src/components/media_control/media_control.js
```diff
@@ -134,7 +134,7 @@ export default class MediaControl extends UIObject {
   }
 
   enable() {
-    if (this.options.chromeless) return
+    if (this.options.chromeless) {return}
```
towerz: This is another undesired result. We should either have no braces at all or have braces with padding.

.eslintrc.js
tchakabam: :+1: 

it wasn't my intention to change the rules at all in fact, i just want to replicate the same with eslint.

src/components/media_control/media_control.js
tchakabam: sure, ideally there should be nothing to fix in the code.

but i think that braces are a good thing. but with padding, i more than agree :+1:

src/components/core/core.js
tchakabam: :+1:

.eslintrc.js
tchakabam: thanks for the help!

tchakabam:
Great, thanks for the feedback guys

Initially I justed wanted to add eslint `--fix` option, but then I gave myself more work it seems. 

Definitely not the most important problem we have, but let's finish what we started ;)

.eslintrc.js
tchakabam: Ok, added that, but it only is for object litterals.

for `if` and other flow control statements you need "curly" rule and in our case the "multi-or-nest" option: https://eslint.org/docs/rules/curly#multi-or-nest

That is what comes most close to the current code style (but it will enforce *not* using braces when there is only a single statement in an if or else, but enforces to nest it)

tchakabam:
@towerz Let me know if this is what you want to go with. 

Additional rule to what you requested: https://eslint.org/docs/rules/curly#multi-or-nest (fits best current code-style and enforces removal of currently existing unnecessary braces around single-line if statements).

We could also add linting to all the config js file in the project root.

tchakabam:
Note: We could also have a sub-config only for the tests.

.eslintrc.js
```diff
@@ -44,6 +44,13 @@ module.exports = {
             "error",
             "never"
         ],
-        "no-var": "error"
+        "no-var": "error",
+        // part of Node Style-guide but ignored
+        "max-len": 0,
+        "max-statements": 0,
+        "curly": ["error", "multi-or-nest"],
```
towerz: @tchakabam I think it's better to use the [consistent](https://eslint.org/docs/rules/curly#consistent) style, instead of `multi-or-nest`. This way we prevent using brace-less statements chained to brace blocks, such as `if/else`.

tchakabam: afaiu `consistent` is in addition to any of the `multi..` options (`"When using any of the multi* options, you can add an option to enforce all bodies of a if, else if and else chain to be with or without braces."`)

let's try that! :)

tchakabam:
@towerz had to add block-spacing rule in order to avoid `if () {foo()}`. lgtm now. 

again, my aim wasn't to fix any inconsistencies in the first place but just unify the toolchain to latest eslint and its ruleset. hope that this will help us to have even nicer code in the future.

this is a nice style, i like it. we should probably have it enforced in Hls.js as well :)

towerz:
Thanks, @tchakabam. This was a great addition to our project and I think it will help us to have much nicer code in the future.

PR #1554, opened 2017-12-22
Adaptive media API and playback base class and extended player config

tchakabam:
Clappr API for adaptive media. 

This is still in ideation state, happy to have an open debate around it! :smile: 

More infos next year ;)


src/base/adaptive_playback.js
```diff
@@ -0,0 +1,102 @@
+import Playback from './playback'
+
+/**
+ * @typedef {Function} AdaptiveMediaActivatorFunction
+ * @function
+ * @param {Boolean} scheduleActivity Enable/disable activity, switch to a different media representation
+ * @param {Boolean} immediateFlush (default = false) Immediate switching, flushes playout buffer
+ * @returns {Boolean} Wether the switch request could be processed
```
tjenkinson: Wether => whether :)

src/base/adaptive_playback.js
```diff
@@ -0,0 +1,102 @@
...
+ * @typedef ClosedCaptionOption
+ * @class
+ * @property {String} id
+ * @property {Boolean} active
+ * @property {String} language
+ * @property {String[]} roles
+ * @member {AdaptiveMediaActivatorFunction} setActive
+ */
+
+export default class AdaptivePlayback extends Playback {
```
tjenkinson: I'm not sure if 'AdaptivePlayback' is the best name given this contains audio options and closed captions?

towerz: I think these could be called MediaOptions. That's something like we're doing in the iOS/Android native versions of Clappr.

src/components/player.js
```diff
@@ -148,6 +148,26 @@ export default class Player extends BaseObject {
...
+   * @param {Number} [options.maxAdaptiveBitrate]
+   * Limits the streamed bitrate (for adaptive media-engines in underlying playback implementations)
+   * @param {Object} [options.maxAdaptiveVideoDimensions]
+   * Limits the video dimensions in adaptive media-engines. Should be a literal object with `height` and `width`.
+   * @param {Boolean}[options.enableAutomaticABR] **default**: `true`
+   * Allows to enable/disable automatic bitrate switching in adaptive media-engines
+   * @param {String} [options.preferredTextLanguage] **default**: `'pt-BR'`
+   * Allows to set a preferred text language, that may be enabled by the media-engine if available.
+   * @param {String} [options.preferredAudioLanguage] **default**: `'pt-BR'`
+   * Allows to set a preferred audio language, that may be enabled by the media-engine if available.
```
towerz: About these parameters, I suggest we keep them inside the playback options. We're avoiding the addition of top level parameters to keep the documentation simpler and to add scope to the options themselves.

src/base/adaptive_playback.js
tchakabam: @tjenkinson Adaptive is not only about (video) quality, but generally having content that adapts to the viewers needs/choices.

src/components/player.js
tchakabam: makes sense :+1:

tchakabam:
@towerz Let's say that I didn't want to force anyone to implement these methods in any `Playback` which is why I wanted to rather make it a mix-in for composition of playbacks that are "adaptive". Like many Playback plugins will not need this stuff in fact. 

Well we could keep the mix-in code here still in a seperate file/module, even if we eventually decide to always have Playback extend from it, right?

tchakabam:
@towerz This isn't about adding "common functions" to the Playback, but to specify some generic interface  to deal with adaptive media - meaning selection of audio/video tracks, qualities, codecs, and subtitles languages etc. It needs to be very generic. Shaka API is a good example of how to achieve that, but it has it's disadvantages as well. We are having a discussion about that on the `DashShakaPlayback` repo as well in fact.

tchakabam:
about the `MediaOptions` you suggested, would you say that we should have one such type instead of the `AudioOption` and `ClosedCaptionOptions` here? 

It would be possible to collapse all of it together, and add a `type`, to what it actually is (audio/video/..), but can you convince me of the advantage?

src/components/player.js
tchakabam: it's a good idea to hiearchise the config API in this way.

but where are the "playback options" documented?

src/components/player.js
```diff
@@ -98,6 +98,48 @@ export default class Player extends BaseObject {
     }
   }
 
+  /**
+   * @typedef {Object} PlaybackConfig
```
tchakabam: @towerz I have created a definition of `PlaybackConfig` the way it is already used in HTML5VideoElement - seems we are already inherently using this but it's not actually documented anywhere.

As this could be an "official" API increment, we will be able to drop the support for the deprecated properties in Player options?

Is this the direction we could go?

tchakabam:
@towerz or anyone, do you want to weigh in on this?

One reason why i want to get this advancing is that currently Levels-Plugin is using undocumented and unspecific API. i would like it to work well as much with Hls.js and with Shaka, and for that I first would like to have a proper API in Clappr to actually do adaptive switching.

src/components/player.js
towerz: @tchakabam I think it's the right direction. I'm not sure if we can drop the support immediately, but we can add a deprecation notice, just to have a smoother transition.

towerz:
@tchakabam I agree with you. There should be a common interface for that. The real problem is to add some breaking changes without planning in advance for a smooth transition.

Instead of doing that, I prefer adding some warning about the deprecation of the "older" API (although it's not documented), just to allow users to refrain from updating to a newer version, but with the possibility of breaking their player instance.

towerz:
Another thing I'd like to keep is that the playback-specific options which are not specified in the `PlaybackConfig` interface should be passed directly to the underlying playback implementation, for example hls.js and Shaka.

tchakabam:
@towerz Yeah totally agree with you, it shouldn't be a breaking API change, and I don't think it wil be necessary to break anything at all :) Playback implementations can transitions smoothly to use the new interface and settings, they will be an addition for now only.

So of course let's keep these in the documentation, and add a deprecated flag  :+1: 

I guess these are just things we need to polish to make sure it's all well set out and documented.

So generally on the proposition of the extended interface (`AdaptivePlayback`), should we move on with this.

What are the todos? :) Apart adding back the removed fields from the config jsdoc?

I guess we need to make sure that these are foundations that we can extend on later and that will allow realizing all future needs, and that is straight-forward to implement and to use.

src/base/adaptive_playback.js
tchakabam: @towerz To what did you refer exactly when stating the `MediaOptions` in iOS/Android Clappr? :)

towerz: @tchakabam actually I was talking about audio/subtitle/etc. But we've already discussed about those options having different interfaces could be beneficial, so nevermind.

towerz:
@tchakabam I think we can move on with the `AdaptivePlayback` interface. Let's just keep the deprecated parameters/properties/methods in the documentation, and proceed with the changes in the existing playbacks without breaking anything.

I agree with you, we can evolve these interfaces later, with the benefit of having a documented common interface.

tchakabam:
@towerz Great to hear that! We'll move on with this asap :) :+1:

tchakabam:
@towerz So far so good in terms of interfaces / docs. We could merge this and try next step to move the Hls.js or Shaka plugins to be driven through this new API.

I am imagining a more powerful demo for adaptive streaming where we would combine usage of LevelPlugin (and others) (which should also use the API spec'd out here) with the various adaptive-enabled Playback plugins (Shaka and Hls.js) to drive all the various things through one API (a/v renditions and text tracks etc).

But for moving on on that, let's first merge this one here?

towerz:
@tchakabam I agree, I think we should move forward on this. I'll merge this now.

tchakabam:
Hey @towerz, just getting back to you on this, it's been a bit slowish here, but we have a real need 

now to make different playback plugins work behind one specified adpative media API :) So soon there will be also real results, not just a scaffold.

So just to let you know, I haven't forgot about this. And planning to do some real progress on this soon!

First of all we'll move the Shaka plugin towards this new API, and then hopefully Hls.js as well.

It will be nice to be able to integrate them easily with tracks/level switching plugins and so on :)

PR #1555, opened 2017-12-22
add editorconf

tchakabam:
Editorconf. Useful to me, probably/hopefully to others too? :) 

http://editorconfig.org/

Let me know if there are things to add/change or if something conflicts with the lint rules.

:beers:

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -14,9 +14,7 @@ indent_size = 2
 trim_trailing_whitespace = true
 
 [*.md]
-trim_trailing_whitespace = true
+# add Markdown specifics if needed
 
-[{package.json}]
-indent_style = space
-indent_size = 2
-insert_final_newline = true
+[*json]
+# add JSON specifics if needed
--- a/.editorconfig
+++ b/.editorconfig
@@ -14,8 +14,9 @@ indent_size = 2
 trim_trailing_whitespace = true
 
 [*.md]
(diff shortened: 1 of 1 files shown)


.editorconfig
```diff
@@ -0,0 +1,21 @@
...
+end_of_line = lf
+insert_final_newline = true
+
+indent_style = space
+indent_size = 2
+
+trim_trailing_whitespace = true
+
+[*.md]
+trim_trailing_whitespace = false
```
towerz: Is there any reason we can't trim whitespaces on `.md` files? If not, I'd rather keep trimming on.

.editorconfig
```diff
@@ -0,0 +1,21 @@
...
+
+[{package.json}]
+indent_style = space
+indent_size = 2
```
towerz: Just to remove some annoying tools complaints, I think it's better to add a new line at the end of the file.

.editorconfig
tchakabam: yes totally agree in fact

tchakabam:
@towerz The config is incremental, it applies matches in order of occurence to files. So I was able to transform your suggestions by boiling it down a bit. Seems what we need is really simple. Let me know if that's it.

towerz:
That's great, @tchakabam. Thanks!

PR #1558, opened 2018-01-05
Lint project root files

tchakabam:
Just a tiny change to have lint scripts also take care of all the js files at project root.

We need to apply fixes there as well then.


towerz:
I'll merge this and fix the lint errors right away since I need to release another version. The last one was broken, due to missing files in the `dist` folder

towerz:
Thanks, @tchakabam!

tchakabam:
Awesome, thank you! :+1:

PR #1561, opened 2018-01-10
Allow configure new events

vagnervjs:
In the current implementation it's only possible to define events callbacks on player constructor.

This PR changes will:

- Allow set new events on configure method;
- Remove previous option event listeners when configure is called with a net set of events;
- Add tests for register option events listeners;


kslimani:
@vagnervjs : would be nice also if you could add behaviour explanation in documentation. (_ie: explain cases where event is added, overide or removed,.._).

Issue #1563, opened 2018-01-24
onReady event not fired when source is an m3u8


marbleman:
**Browser**: Chrome  63.0.3239.132 , Firefox Version 57.0.4 and all others such as Safari etc.)

**OS**: Windows (7, 8 and 10)

**Clappr Version**: 0.2.86 (as of today)

**Steps to reproduce**:
 Use  [http://clappr.io/demo/](url)
and add onReady event:
`var playerElement = document.getElementById("player-wrapper");

var player = new Clappr.Player({
  source: 'http://clappr.io/highline.mp4',
  poster: 'http://clappr.io/poster.png',
  mute: true,
  height: 360,
  width: 640,
  events: {
    onReady: function() {alert('onReady triggered')}
 }  
});

player.attachTo(playerElement);`

When clicking "run", the onReady event will fire.

Now modify the source to 
`https://bitdash-a.akamaihd.net/content/sintel/hls/playlist.m3u8`

onReady will not fire unless you hit the play button.
It does not matter whether m3u8 contains multiple streams or just a single stream.

For my understanding onReady is supposed to trigger when player setup has completed. Its behaviour should not depend on the type of source. Or am I missing something important?

marbleman:
Addition: It's getting more confusing...

On Chrome 63.0.3239.132 onReady will fire for an m3u8 but _only_ if 'poster' is NOT an empty string, null or missing in the config object. Firefox and Safari still do not fire onReady in any case...

leandromoreira:
@marbleman I just tested on my Firefox 57.0.4 (MacOS) and everything works as expected and I tried with an mp4 and an hls.

![screen shot 2018-01-23 at 10 56 45 pm](https://user-images.githubusercontent.com/55913/35308860-ce2423ec-0090-11e8-9368-7a302559641e.png)

marbleman:
And onReady fires immediately when you hit the "Run" button that evaluates the code? For me it does not! Not even on Firefox on macOS... It just fires when I click the video to play (at the same time as onPlay)... 

Results as:
N: onReady fires not on player ready but only together with onPlay 
Y: onReady fires correctly when player is ready (and still idle... so independent of onPlay)

Tested so far on:
Windows 7:
   - Firefox 57.0.4: progressive: **Y**, m3u8: **N**
   - Chrome 63.0.3239.132: progresssive: **Y** , m3u8: **Y** (but only if poster paremeter is not empty)
[...]

PR #1578, opened 2018-02-21
Suggested Fix: onReady event firing with m3u8

wahajdar:
Suggested fix for #1563

### DIFF
--- a/src/playbacks/hls/hls.js
+++ b/src/playbacks/hls/hls.js
@@ -159,13 +159,14 @@ export default class HLS extends HTML5VideoPlayback {
     this._hls.on(HLSJS.Events.SUBTITLE_TRACK_LOADED, (evt, data) => this._onSubtitleLoaded(evt, data))
     this._hls.on(HLSJS.Events.SUBTITLE_TRACKS_UPDATED, () => this._ccTracksUpdated = true)
     this._hls.attachMedia(this.el)
-    this._ready()
+  }
+
+  render() {
+    this._ready();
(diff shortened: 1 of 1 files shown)


leandromoreira:
Thanks, @wahajdar did you see the errors on [CI](https://travis-ci.org/clappr/clappr/builds/344230077?utm_source=github_status&utm_medium=notification)?

wahajdar:
Apologies, let me consider. 
Thanks @leandromoreira

wahajdar:
@leandromoreira , can you please guide me how could I fix coverage check.
Will be grateful, thanks

towerz:
@wahajdar Thanks, this is a great addition. The ready event should be reliable to allow code to run as soon as the playback can accept calls. Otherwise, we could end with a number of issues or messy code handling different types of playbacks.

About the code coverage, don't worry for now. @joaopaulovieira is working on a test framework change we want to do so just not breaking the current tests is ok.

I'm working on an FSM implementation for the base playback, in order to allow all the playbacks to share the same state semantic. Basically, the playbacks will all implement the state machine described [here](https://github.com/clappr/clappr-docs/wiki/Architecture#playback)

wahajdar:
@towerz thank you for letting me contribute. Thanks for clarifying certain things here, will follow clappr architecture docs for better understanding.

poshe:
class **HLS** extends **HTML5VideoPlayback** and they both have the same `_ready()` function and they both call it in their `render()` functions. Is that OK?

PR #1580, opened 2018-02-22
Create PLAYBACK_SEEK event

### LINKED ISSUE #1162: Question : any reason there is no playback "seek" event ?
As far i have dig in source code, the player seek event comes from container seek event which is trigger in container [seek()](https://github.com/clappr/clappr/blob/master/src/components/container/container.js#L302) method.

I am working on player event tracking plugins (Google Analytics, etc..) and this is problematic in some case, for example, iOS device in fullscreen where Clappr media controls cannot be rendered. (_seek event is never trigger_)

What about a `Events.PLAYBACK_SEEK` and a `Events.PLAYBACK_SEEKED` events ?

EDIT: by "problematic", i mean i can't rely on binding container events.

joaopaulovieira:
resolve #1162 creating a new event `PLAYBACK_SEEK`
Also, fixes a test for HLS playback spec.

EDIT: Add test for new event too.

### DIFF
--- a/src/base/events.js
+++ b/src/base/events.js
@@ -484,6 +484,12 @@ Events.PLAYBACK_PLAY = 'playback:play'
  */
 Events.PLAYBACK_PAUSE = 'playback:pause'
 /**
+ * Fired when the media for a playback is seeking.
+ *
+ * @event PLAYBACK_SEEK
+ */
+Events.PLAYBACK_SEEK = 'playback:seek'
+/**
  * Fired when the media for a playback is seeked.
  *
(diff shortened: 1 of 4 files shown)


leandromoreira:
Can you try to add a test case for this event? Here's [a similar spec](https://github.com/clappr/clappr/blob/master/test/playbacks/html5_video_spec.js#L51).

joaopaulovieira:
Yeah! I was looking for some similar test, but in the `events_spec.js` :sweat_smile:

PR #1593, opened 2018-03-14
enable MediaControl configuration after player creation

markmatney:
Summary of changes:
- add `MEDIACONTROL_OPTIONS_CHANGE` event for consistency with `Container`'s events
- add `configure` method to `MediaControl` and call it from `Core.configure` (TODO: check whether any `MediaControl`-specific options were actually changed before calling it, to prevent erroneous event triggering)
- add test

### DIFF
--- a/src/components/core/core.js
+++ b/src/components/core/core.js
@@ -346,6 +346,7 @@ export default class Core extends UIObject {
     this.containers.forEach((container) => {
       container.configure(this.options)
     })
+    this.mediaControl.configure(this.options)
   }
 
   appendToParent() {
--- a/src/components/media_control/media_control.js
+++ b/src/components/media_control/media_control.js
@@ -644,6 +644,16 @@ export default class MediaControl extends UIObject {
     this.unbindKeyEvents()
   }
 
+  /**
+   * enables to configure the media control after its creation
+   * @method configure
+   * @param {Object} options all the options to change in form of a javascript object
+   */
+  configure(options) {
+    this._options = $.extend(this._options, options)
+    this.trigger(Events.MEDIACONTROL_OPTIONS_CHANGE)
+  }
+
   render() {
     const timeout = 1000
     this.$el.html(this.template({ settings: this.settings }))
--- a/src/base/events.js
+++ b/src/base/events.js
@@ -795,3 +795,9 @@ Events.MEDIACONTROL_NOTPLAYING = 'mediacontrol:notplaying'
  * @event MEDIACONTROL_CONTAINERCHANGED
  */
 Events.MEDIACONTROL_CONTAINERCHANGED = 'mediacontrol:containerchanged'
+/**
+ * Fired when the options were changed for the mediacontrol
+ *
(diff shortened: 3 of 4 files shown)


flavioribeiro:
lgtm!

src/components/core/core.js
```diff
@@ -346,6 +346,7 @@ export default class Core extends UIObject {
     this.containers.forEach((container) => {
       container.configure(this.options)
     })
+    this.mediaControl.configure(this.options)
```
vagnervjs: Would be nice to check if  `configure` is a method, because some Clappr use cases can add an external Media Control without this new configure method, and will get an 
Uncaught TypeError: `mediaControl.configure is not a function`

src/components/media_control/media_control.js
```diff
@@ -644,6 +644,16 @@ export default class MediaControl extends UIObject {
     this.unbindKeyEvents()
   }
 
+  /**
+   * enables to configure the media control after its creation
+   * @method configure
+   * @param {Object} options all the options to change in form of a javascript object
+   */
+  configure(options) {
+    this._options = $.extend(this._options, options)
```
vagnervjs: Can we clone only Media Control options, to avoid duplication of unnecessary options inside Media Control? 
Something like:  `this._options = $.extend(this._options, options.mediacontrol)`

src/components/core/core.js
leandromoreira: Shouldn't all the (configurable) components answer to this method? (even if it's to be an empty implementation)

leandromoreira: Once we define that a refresh on any configuration will trigger a calling to all configurable components with the newConfig, this might be helpful.

markmatney: @vagnervjs looks like I should follow [this style](https://github.com/clappr/clappr/blob/147867b33c9bcd1268aff6505bbbdb33d2a01e62/src/base/polyfills.js#L27). Should I open a new PR with these changes?

src/components/media_control/media_control.js
markmatney: I agree, however I think I would need help since I am not that familiar with this codebase and it appears that not all `MediaControl`-specific options are namespaced under `options.mediacontrol` (for example, `options.hideMediaControl`, and presumably also `options.hideVolumeBar`).

Likewise, it seems like [`Container.configure`](https://github.com/clappr/clappr/blob/147867b33c9bcd1268aff6505bbbdb33d2a01e62/src/components/container/container.js#L463) should also only be called if container-specific options are changed. Perhaps [`Core.configure`](https://github.com/clappr/clappr/blob/147867b33c9bcd1268aff6505bbbdb33d2a01e62/src/components/core/core.js#L338) should look like this:
```javascript
...
this.containers.forEach((container) => {
    if (containerSpecificOptionsChanged) {
        container.configure(this.options)
    }
})
if (mediaControlSpecificOptionsChanged) {
    this.mediaControl.configure(this.options)
}
...
```

PR #1594, opened 2018-03-14
Update documentation (flv playback plugin)

andrefilimono:


leandromoreira:
Thank you very much @andrefilimono

Issue #626, opened 2015-11-02
RFC: Add support to manage audio/video/text tracks at Playback interface


leandromoreira:
While I'm creating the dash playback I'm seeing a lot of opportunities to clappr. :) Dash/HLS offers multiple audio/video/text tracks.

I think it might be useful to add these capabilities to our `Playback`.

I also think that @towerz is working to open source our solution to Adobe DRM (which also handles multiples audios...)

**This is very inspired by [shaka](https://shaka-player-demo.appspot.com/docs/shaka.player.Player.html).**
## Propose

We can treat a `track` as a simple abstraction regardless of this type (audio, video or text)

``` javascript
text = {
  id: 0, //*mandatory: an unique identifier
  active: false, //*mandatory: weather this track is active or not,
  lang: 'pt-BR' //optional: the track's language
}

audio = {
  id: "A",
  active: true,
  lang: 'en-US'
}

video = {
  id: "3D",
  active: false,
  lang: null
}
```

The idea is to support the following operations:
-  `enableTextTrack(enable=true)`  //this is needed, we can mute and we need to be watching at least one video track :P
-  `audioTracks()`
-  `selectAudioTrack(id)`
-  `textTracks()`
-  `selectTextTrack(id)`
-  `videoTracks()`
-  `selectVideoTrack(id)`
### Solution A

**Just append more methods to Playback.**

**Good**: simple adhoc, **Bad**: not all playbacks needs to know manage audio/video/text tracks
### Solution B

**Create a new class like `AVTTrackHandler` and compose Playback (aka `get avtTrackerHandler`).**

**Good**: not so coupled, **Bad**: force us to always to know and use the compositon `playback.avtTrackerHandler.selectVideoTrack`
### Solution C

**Create a new sub-class like `PlaybackWithTrackHandler` and extend from it.**

**Good**: not so coupled, **Bad**: yet another layer of heritance :-1: 
### Solution D

????
## Question
- Do we need more fields to represent a track? (maybe `type: 'audio'` but I don't think of any use cases for it)
- What do you think?
- What solution do you like most?

kevin2o:
Was this implemented? I love the player, but I have playlists with multiple audio tracks and I can't find an option to change the track.

leandromoreira:
@kevin2o it was not ;/ but I still think it's a solid feature

kevin2o:
[...]

PR #1602, opened 2018-03-19
Hlsjs adaptive playback

tchakabam:
Basic support for Hls.js playback to implement novel AdaptivePlayback API. 

First PoC for this new interface.

Example:
```
// activates to level 0 (worst quality) with immediate flush (will switch directly)
player.core.getCurrentPlayback().videoQualityLevels[0].setActive(true, true)

// prints array object containing all quality level handles 
console.log(player.core.getCurrentPlayback().videoQualityLevels)

etc ...
```

### DIFF
--- a/src/base/adaptive_playback.js
+++ b/src/base/adaptive_playback.js
@@ -63,7 +63,12 @@ export default class AdaptivePlayback extends Playback {
   /**
    * @param {Boolean} enabled
    */
-  set isAutoAdaptive(enabled) {}
+  set isAutoAdaptive(enabled) {
+    if (this.isAdaptive) {
+      throw new Error('Playback is adaptive but not implemented')
+    }
+    return false
+  }
 
   /**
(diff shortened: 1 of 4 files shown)


leandromoreira:
https://github.com/clappr/clappr/issues/626

tchakabam:
@leandromoreira I think we can implement the methods from your proposal above easily now as common methods of the AdaptivePlayback base-class, with the implementations supporting the generic abstract methods that are a little bit more versatile, or can serve more wide-range cases.

src/base/playback.js
```diff
@@ -242,6 +242,8 @@ Playback.VOD = 'vod'
  * @property AOD
  * @static
  * @type String
+ * @deprecated There is technically no difference with VOD.
+ *             Or: if this case exists, it is inconsistent that there is no difference between audio/video "live".
```
tchakabam: TBD

PR #1603, opened 2018-03-19
wip improve demo default setup

tchakabam:
some ideas tbd to improve the default demo code :)

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -1,5 +1,7 @@
 <html>
   <head>
+    <title>Clappr Demo</title>
+
     <meta http-equiv="Content-Type" content="text/html; charset=utf-8">
     <link rel="stylesheet" href="stylesheets/bootstrap.min.css" type="text/css" media="screen" charset="utf-8">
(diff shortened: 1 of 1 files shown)


public/index.html
```diff
@@ -48,18 +50,29 @@
...
       <div class="sidebar">
-        <div id="editor">var playerElement = document.getElementById("player-wrapper");
+        <div id="editor">
+var clapprDemo = window.clapprDemo = {}
+
+var playerElement = document.getElementById("player-wrapper");
 
 var player = new Clappr.Player({
-  source: 'http://clappr.io/highline.mp4',
+  source: 'https://video-dev.github.io/streams/x36xhzz/x36xhzz.m3u8',
```
tchakabam: this is my wip, we can have the mp4 as well in there :)

But it would be cool to show a bit more in the demo the HLS and DASH integrations of Clappr, also to keep making sure all of this stuff integrates well with all the plugins (like level-switching, captions etc)

tchakabam:
One thing that is bad: The duplicate code between the actual script and the editor for startup.

The editor should pull in the initial setup code from a seperate file, as text XHR, and fill it with that, and then just call the editor run routine internally - don't "special-case" the initial run.

PR #1613, opened 2018-03-28
Feature/error component

jhonatangomes:
To normalize errors and create a pattern to them a new error component is being created. 

Initially only playback errors are being changed to this new format, but errors from other layers can adopt this new format too.

This component will trigger a new error event that contains data that can be useful to, for example, an error screen plugin render information to users.

This new error format contains: 

**code**: the error code
**description**: error description
**level**:  warn or fatal
**origin**: hls, html5, etc
**scope**: playback, container, etc
**raw error**: the error without modification

### DIFF
--- a/src/base/events.js
+++ b/src/base/events.js
@@ -282,7 +282,7 @@ Events.PLAYER_SEEK = 'seek'
  * @event PLAYER_ERROR
  * @param {Object} error the error
  */
-Events.PLAYER_ERROR = 'error'
+Events.PLAYER_ERROR = 'playererror'
 /**
  * Fired when the time is updated on player
  *
--- a/src/base/events.js
+++ b/src/base/events.js
@@ -289,6 +289,18 @@ Events.PLAYER_ERROR = 'playererror'
  * @event ERROR
  * @param {Object} error
  * the error with the following format `{code, description, level, raw, origin, scope}`
+ * @param {String} [options.code]
+ * error's code: code to identify error in the following format: origin:code
+ * @param {String} [options.description]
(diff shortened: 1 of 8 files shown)


src/base/events.js
```diff
@@ -282,7 +282,15 @@ Events.PLAYER_SEEK = 'seek'
  * @event PLAYER_ERROR
  * @param {Object} error the error
  */
-Events.PLAYER_ERROR = 'error'
+Events.PLAYER_ERROR = 'playererror'
+/**
+ * Fired when there is an error
+ *
+ * @event ERROR
+ * @param {Object} error
```
leandromoreira: Can you describe/document the proposed error object here? Like we do at [Player's constructor](https://github.com/jhonatangomes/clappr/blob/c5e5fcf174c9d60135750608d4e3af5c4f4aa45e/src/components/player.js#L152-L160)?

jhonatangomes: yes, done.

src/playbacks/html5_video/html5_video.js
```diff
@@ -383,7 +383,15 @@ export default class HTML5Video extends Playback {
   }
 
   _onError() {
-    this.trigger(Events.PLAYBACK_ERROR, this.el.error, this.name)
+    const { code, message } = this.el.error
```
otmjka: there will be an error if `this.el.error` is `null`

jhonatangomes: This PR is going on at https://github.com/clappr/clappr/pull/1614

Issue #1108, opened 2016-08-05
Add default error screen


towerz:
When a playback error occurs, like the one reported in #1105

PR #1614, opened 2018-03-28
Feature/error component

jhonatangomes:
To normalize errors and create a pattern to them a new error component is being created.

Initially only playback errors are being changed to this new format, but errors from other layers can adopt this new format too.

This component will trigger a new error event that contains data that can be useful to, for example, an error screen plugin render information to users.

This new error format contains:

**code**: the error code
**description**: error description
**level**: warn or fatal
**origin**: hls, html5, etc
**scope**: playback, container, etc
**raw error**: the error without modification

This is the first step to create our error screen #1108

### DIFF
--- a/src/components/error/error.js
+++ b/src/components/error/error.js
@@ -36,7 +36,7 @@ class PlayerError extends BaseObject {
    */
   error(err) {
     if (!this.core) {
-      Log.warn(this.name, 'Core is not setted. Error: ', err)
+      Log.warn(this.name, 'Core is not set. Error: ', err)
       return
     }
     this.core.trigger(Events.ERROR, err)
--- a/src/components/error/error.js
+++ b/src/components/error/error.js
@@ -19,7 +19,8 @@ class PlayerError extends BaseObject {
   get Levels() {
     return {
       FATAL: 'FATAL',
-      WARN: 'WARN'
+      WARN: 'WARN',
+      INFO: 'INFO',
     }
   }
 
--- a/src/components/error/error.js
+++ b/src/components/error/error.js
@@ -1,0 +1,46 @@
+import Events from '../../base/events'
+import BaseObject from '../../base/base_object'
+
+/**
+ * The PlayerError is responsible to receive and propagate errors.
+ * @class PlayerError
+ * @constructor
+ * @extends BaseObject
+ * @module components
+ */
+class PlayerError extends BaseObject {
+  get name() { return 'error' }
+
+  /**
+   * @property Levels
+   * @type {Object} object with error levels
+   */
+  get Levels() {
+    return {
+      FATAL: 'FATAL',
+      WARN: 'WARN'
+    }
+  }
+
+  constructor(options={}) {
+    super(options)
+  }
+
+  /**
+   * creates and trigger an error.
+   * @method error
+   * @param {Object} err should be an object with code, description, level, origin, scope and raw error.
+   */
+  error(err) {
+    this.core.trigger(Events.ERROR, err)
+  }
+
+  setCore(core) {
+    this.core = core
+  }
+
+}
+
+const playerError = new PlayerError()
+
+export default playerError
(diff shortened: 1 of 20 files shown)


jhonatangomes:
Some changes were requested by @leandromoreira on this pull request https://github.com/clappr/clappr/pull/1613 that was closed because of the branch was properly renamed.

test/components/error_spec.js
```diff
@@ -0,0 +1,40 @@
...
+      code: 'test_01',
+      description: 'test error',
+      level: PlayerError.Levels.FATAL,
+      origin: 'test',
+      scope: 'it',
+      raw: {},
+    }
+  })
+
+  it('trigger ERROR with error data when error method is called', function() {
```
leandromoreira: `it triggers the` or we could:

```javascript
describe("when error method is called", function(){
    it("triggers the event ERROR", ...
})
``` 

The fact of ERRO being fired with error data is already expected.

test/components/error_spec.js
```diff
@@ -0,0 +1,40 @@
...
+      code: 'test_01',
+      description: 'test error',
+      level: PlayerError.Levels.FATAL,
+      origin: 'test',
+      scope: 'it',
+      raw: {},
+    }))
+  })
+
+  it('does not trigger ERROR when core is not setted', function() {
```
leandromoreira: Here I also think we can use context:

```
describe("when core is not ready", ....)

File: test/playbacks/hls_spec.js [hunk 3]
```diff
@@ -80,16 +80,23 @@ describe('HLS playback', () => {
     })
   })
 
-  it('should trigger a playback error if source load failed', function(done) {
-    this.timeout(5000)
+  it('should trigger a playback error if source load failed', function() {
```
leandromoreira: Here we could use the usual phrase but I'd vote to remove `should` :P

jhonatangomes: Yes, I agree. But the tests are with lot of should on it messages. I fixed the test but didn't change its message. I think a refactor to change all these messages is better.

test/components/error_spec.js
jhonatangomes: fixed on e6917a0eeb6bacfe6985cefc5a147bbcc12681cb and 8404cf9a80e2273e3879e426597d5f70a1960294

test/components/error_spec.js
jhonatangomes: improved on e6917a0eeb6bacfe6985cefc5a147bbcc12681cb and 8404cf9a80e2273e3879e426597d5f70a1960294

src/components/error/error.js
```diff
@@ -0,0 +1,51 @@
...
+    this.core.trigger(Events.ERROR, err)
+  }
+
+  setCore(core) {
+    this.core = core
+  }
+
+}
+
+const playerError = new PlayerError()
```
towerz: I wouldn't recommend using a singleton approach for this component. Since there could be more than a single player in a given context, an error would always propagate to the last instance which called `setCore`.

towerz: Instead of doing that, we could create the error component when the core is instantiated, perhaps in the `CoreFactory` or in the `Core` itself.

test/playbacks/hls_spec.js
towerz: I agree with removing `should`. Using the present tense would make the test declarations more descriptive.

test/base/playback_spec.js
```diff
@@ -33,4 +34,62 @@ describe('Playback', function() {
...
+
+  it('creates a code error on the following format: name:code', () => {
+    this.basePlayback.name = 'test'
+    const error = { code: '42' }
+    const errorData = this.basePlayback.createError(error)
+
+    expect(errorData.code).to.deep.equal(`${this.basePlayback.name}:${error.code}`)
+  })
+
+  it('default error level equals to FATAL', () => {
```
towerz: Let's try to make `it` descriptions more like a single sentence. Something like `it("triggers a FATAL error by default")`.

src/components/error/error.js
jhonatangomes: fixed on cb5fe1fbce139bca7d01f5c48c8d6ba63e2cae31

PR #1619, opened 2018-04-03
Refactor MediaControl as a UICorePlugin

### LINKED ISSUE #1606: Refactor MediaControl as a UICorePlugin
Instead of being a separate Core component with access to the Core scope, the MediaControl should be just another UICorePlugin instance. 

This way it could be easier to replace it entirely, as well as removing the constraint of having a mandatory MediaControl in the player.

joaopaulovieira:
Closes #1606 

TL;DR
====
The media control component has become a `UICorePlugin` and the core component now defines what the current container is, so it's not necessary to have the media control reference on the plugins that need to know which container is currently active.

Changes
======
* MediaControl component now is a `UICorePlugin`;
* The `core` component now use `MediaControl` plugin reference or a `dummyMediaControl` for plugins that need the media control reference;
* The name of the media control plugin will have to be `media_control`;
* The `core` component has a `activeContainer` get/set to manage the current container;
* New event `CORE_ACTIVE_CONTAINER_CHANGED` triggered when the `activeContainer` is set;
* Plugins that need to know what the current container no longer need to have the media control reference to listen the `MEDIACONTROL_CONTAINERCHANGED` event;
* Media control tests adapted to the new structure;


src/components/core/core.js
```diff
@@ -63,14 +62,30 @@ export default class Core extends UIObject {
     return this.getPlugin('strings') || { t: (key) => key }
   }
 
+  get mediaControl() {
```
leandromoreira: I think we could document all the public interface of this component since it'll be very useful for players developers.

src/components/core/core.js
```diff
@@ -247,7 +239,8 @@ export default class Core extends UIObject {
     containers.map(this.appendContainer.bind(this))
     this.trigger(Events.CORE_CONTAINERS_CREATED)
     this.renderContainers()
-    this.setupMediaControl(this.getCurrentContainer())
+    this.activeContainer = containers[0]
```
leandromoreira: Is it safe to assume that `0` will always the active container? where does this policy is enforced?

src/components/core/core.js
towerz: @leandromoreira this "public" interface is only for compatibility purposes, in order to not break existing plugins. It will be removed in a later version, so it's already "deprecated"

src/components/core/core.js
```diff
@@ -63,14 +62,30 @@ export default class Core extends UIObject {
...
 
+  get mediaControl() {
+    return this.getPlugin('media_control') || this.dummyMediaControl
+  }
+
+  get dummyMediaControl() {
+    return new UICorePlugin(this)
+  }
+
+  get activeContainer() {
```
towerz: The `activeContainer` is a new interface, create to replace the `getCurrentContainer` function. This one should be documented.

src/components/core/core.js
```diff
@@ -184,19 +196,17 @@ export default class Core extends UIObject {
     this.options.mimeType = mimeType
     sources = sources && sources.constructor === Array ? sources : [sources]
     this.containers.forEach((container) => container.destroy())
-    this.mediaControl.container = null
+    this.mediaControl && (this.mediaControl.container = null)
```
towerz: I think this assignment doesn't make sense anymore since the `MediaControl` now uses the `activeContainer` related events

src/components/core/core.js
```diff
@@ -184,19 +196,17 @@ export default class Core extends UIObject {
...
+    this.prepareContainers()
   }
 
   destroy() {
     this.disableResizeObserver()
     this.containers.forEach((container) => container.destroy())
     this.plugins.forEach((plugin) => plugin.destroy())
     this.$el.remove()
-    this.mediaControl.destroy()
+    this.mediaControl && this.mediaControl.destroy()
```
towerz: This `destroy` call doesn't make sense anymore since the `MediaControl` is a `UICorePlugin` and the plugins are destroyed in the previous `forEach` loop.

src/components/core/core.js
towerz: @leandromoreira this is a convention already in place. If you look at the previous `getCurrentContainer` implementation, we always assumed that the current container was the first container unless otherwise stated.

Since the containers are only created when there's at least a single source, it's safe to assume that the first container should be initially the active one. If there is more than one source, it's still valid to assume that, since the media control should refer to a single container to control it. Also, when there's more than one source, we should assume that a plugin would take control over what container is currently active, for example, a picture-in-picture plugin or a grid plugin.

src/components/core/core.js
```diff
@@ -264,70 +257,50 @@ export default class Core extends UIObject {
...
-      return new MediaControl(options).render()
-
-  }
-
   getCurrentContainer() {
-    if (!this.mediaControl || !this.mediaControl.container)
-      return this.containers[0]
-
-    return this.mediaControl.container
+    return this.activeContainer
```
towerz: I think we could mark this function as "deprecated",  to discourage using it in favor of using `activeContainer`.

src/components/core/core.js
```diff
@@ -264,70 +257,50 @@ export default class Core extends UIObject {
...
   }
 
   getCurrentPlayback() {
-    const container = this.getCurrentContainer()
-    return container && container.playback
+    return this.activeContainer && this.activeContainer.playback
```
towerz: I think we could mark this function as "deprecated" as well and create an `activePlayback` getter.

PR #1629, opened 2018-04-17
Improve error

jhonatangomes:
This PR adds a new option to prevent overwrite of error code, remove possibly translated wrong messages and does not append container to error object.


bikegriffith:
While you're in here, might be good to clean up the english here:
https://github.com/clappr/clappr/pull/1629/files#diff-60ed5c9c4e5f3456d3cff75bbad73352R40

It would be better to say `"Could not play video."`, as reproducing a video doesn't make much sense in US English.

PR #1636, opened 2018-05-07
Upgrade browser info

joaopaulovieira:
Adds new data on `Browser` Object:
```
data: {
    name: 'browser name';
    group: 'browser group';
    fullVersion: 'version of the browser';
    majorVersion: 'major version of the browser';
    minorVersion: 'minor version of the browser';
},
os: {
    name: 'operational system name;
    group: 'operational system group';
    fullVersion: 'version of the operational system';
    majorVersion: 'major version of the operational system';
    minorVersion: 'minor version of the operational system';
},
viewport: {
    width: 'device screen width';
    height: 'device screen height';
    orientation: 'landscape or portrait' (only mobile);
},
device: 'device name';
```


src/components/browser/index.js
```diff
@@ -0,0 +1,217 @@
+// Copyright 2014 Globo.com Player authors. All rights reserved.
```
vagnervjs: It would be better to keep the `index.js` file just as an entry point to the module.
I suggest creating a `browser.js` file with this code, and `index.js` just load and export the `browser.js` content

vagnervjs: Also, we can take advantage to remove these old comment header :)

src/components/browser/index.js
```diff
@@ -0,0 +1,217 @@
...
+Browser.isAndroid = /Android/i.test(navigator.userAgent)
+Browser.isWindowsPhone = /Windows Phone/i.test(navigator.userAgent)
+Browser.isWin8App = /MSAppHost/i.test(navigator.userAgent)
+Browser.isWiiU = /WiiU/i.test(navigator.userAgent)
+Browser.isPS4 = /PlayStation 4/i.test(navigator.userAgent)
+Browser.hasLocalstorage = hasLocalstorage()
+Browser.hasFlash = hasFlash()
+
+Browser.name = browserInfo.name
+Browser.version = browserInfo.version
```
vagnervjs: Now, this properties will be redundant with `Browser.data.name` and `Browser.data.fullVersion`. Maybe we should remove them?

joaopaulovieira: For reasons of backward compatibility, it would be better to leave both methods at this time and warn they will be removed in a next version (0.3.0)

vagnervjs: 👍

PR #1640, opened 2018-05-13
Extract event constants to their own bundle

bent0b0x:
**Motivation**
Since the distributed version of clappr is somewhat large, it is nice to load the clappr code asynchronously and not include it in an application bundle. However, it is useful to have the constants used for clappr events accessible at build time.

**Work Done**
- move clappr event constants to their own file
- add a new entrypoint (`src/constants/events.js`) to the `webpack` config so that we get a new output of `dist/events.js`

**Example Usage**
```javascript
import Events from 'clappr/dist/events';

...

player.on(Events.PLAYER_PAUSE, console.log);
```

☝️, now we can subscribe to events even before the entire clappr bundle/player code has been imported/loaded.

Very open to feedback/restructuring/refactoring!


bent0b0x:
I will fix merge conflicts soon

bent0b0x:
I am going to close this for now. I think there is an event-binding syntax that does not require knowing the event constants ahead of time that will help

bent0b0x:
Is the coveralls decrease a problem in this case?

leandromoreira:
No @bent0b0x

towerz:
Hi @bent0b0x, what's the actual use you have for separating the event constants into a distributed module? Are you using Clappr package from NPM only for the constants and then using a hosted version of it?

I'm asking this because you don't need to have a  separate entry point just to expose the constants. You could, for example, add a resolve configuration to load the Events module from source, instead of a build on in the dist folder.

bent0b0x:
Hi @towerz!

Thanks for the comment. 

I would like access to the event constants because I would like to define event handlers immediately in a component that loads the entire clappr asynchronously (via a dynamic import)

I did consider loading from the `src` dir. However that did not see very stable to me; source can easily be reorganized without it being a breaking change. Additionally, source can be included in `.npmignore` (although for clappr right now it is not).

I did think of other ways to solve this, such as not defining the event handlers until clappr has loaded async. Perhaps that would be the most frictionless solution to this specific problem.

towerz:
@bent0b0x indeed, delaying the definition of event handlers would work.

What kind of events are you trying to listen? The `Player` object has already some defined events for binding via options:

```
  //...
  events: {
    onPause: ...,
    onPlay: ...
  },
  //...
})

Have you tried using those already?

About having the player events constants loading asynchronously, it seems like a good idea. I'm not so sure about having a separate bundle for those constants. Perhaps there's a better option to expose them without actually having to load Clappr.

About reorganizing source code, we actually consider changing any public module path a breaking change. Importing from source would not be an issue in this case.

bent0b0x:
Thanks @towerz; I'm going to close as I'm just going to bind the events async.

I'm binding to some `PLAYBACK_XXX` events, so I'm not sure if the config options will work. Either way, I have a workaround!

Issue #1639, opened 2018-05-11
Cannot autoplay on the newest google chrome.


zijian85:
**Browser**: Google Chome 66.0.3359.139  64Bit.

**OS**:  Mac OS 10.10.4

**Clappr Version**: 0.2.83

Hi, today I updated my chrome to the newest. The version is  66.0.3359.139  64Bit.

But cannot autoplay now. I have set the autoplay to true and autoplay on older chrome version.

<img width="425" alt="2018-05-11 12 43 35" src="https://user-images.githubusercontent.com/30042173/39907099-1a6f1518-5519-11e8-8e56-3ab9b36e412a.png">

<img width="829" alt="2018-05-11 12 45 13" src="https://user-images.githubusercontent.com/30042173/39907114-2d3e8282-5519-11e8-8e2d-025daa9ba6c0.png">

kslimani:
@zijian85 i think this is due to new autoplay policy with Chrome version 66.

Also, please note that version 0.2.83 should not be used on any environment ([see releases](https://github.com/clappr/clappr/releases/tag/0.2.83)).

Since Google 66, the autoplay policy seems to be based on a "score" which you can check by typing `chrome://media-engagement` in Chrome adress bar.

You may also get the same issue with Safari version 11.

Please note also on Clappr, the promise exception message has been "silenced" some times ago in  commit 2d2f351.

But since these browsers policy changes, the promise should be used to check if player can autoplay or not. _(for example, using similar mechanisms as [can-autoplay](https://github.com/video-dev/can-autoplay))_.

Wiicast:
When I remove the plugins from my player, it works in Chrome, it seems the plugins is the problem.

kslimani:
Here an exemple to handle autoplay, using can-autoplay :

```html
<html>
  <head>
    <title>My App</title>
    <style>
      .player {
        width: 640px;
        height: 360px;
      }
    </style>
    <script type="text/javascript" src="https://cdn.jsdelivr.net/npm/can-autoplay@3.0.0/build/can-autoplay.min.js"></script>
    <script type="text/javascript" src="https://cdn.jsdelivr.net/npm/clappr@latest/dist/clappr.min.js"></script>
  </head>
  <body>
    <div class="app">
        <div class="player"></div>
    </div>
    <script>
      var player;
      canAutoplay.video().then(function(o) {
        var cap = o.result === true;
        player = new Clappr.Player({
          source: 'http://clappr.io/highline.mp4',
[...]

PR #1644, opened 2018-05-16
feat(playback): handle autoplay policy

kslimani:
This PR is __a proposition__ to handle new autoplay policies introduced with Chrome 66 and Safari 11.
_(but also handle for mobile devices)_. 

__Changes :__

Base Playback class :

* adds `consented` property to base playback (boolean which indicate if playback has user consent)
* adds `autoPlay` and `canAutoPlay` methods

HTML5 Video playback :

* overrides `autoPlay` and `canAutoPlay` methods to handle HTMLMediaElement autoplay policy

Utils :

* add `canAutoPlayMedia` helper method (heavily inspired from [can-autoplay](https://github.com/video-dev/can-autoplay) lib)
* add media module which contains base64 encoded mp4 file. _(i choosed this over blob version because of lower size)_.

Edit: it also add a new `autoPlayTimeout` player constructor option. _(to control autoplay check timeout in milliseconds)_.

__Motivations :__

Currently, autoplay logic is handled inside each playback. This is why i added an `autoPlay` method. (called in constructor, instead of `play` method).

This change allow, for example, in the future, to move the autoplay triggering into core instead of playback constructor.

It should simplify also other playback extending `HTML5Video` to fix this issue. (for example, in dash-shaka-playback plugin, ~~it should only require to change `this.play()` by `this.autoPlay()`~~ [here](https://github.com/clappr/dash-shaka-playback/blob/master/src/clappr-dash-shaka-playback.js#L63).
EDIT: @leandromoreira in fact, in Dash Shaka playback, the line should be removed, since super() is called in constructor.


src/base/playback.js
```diff
@@ -210,6 +221,25 @@ export default class Playback extends UIObject {
...
     this.remove()
   }
+
+  /**
+   * attempt to autoplays the playback.
+   * @method autoPlay
+   */
+  autoPlay() {
+    this.canAutoPlay((result, error) => { // eslint-disable-line no-unused-vars
+      result && this.play()
```
kslimani: Here we could maybe log the error as warning message ?

kslimani:
@leandromoreira : could you please re-run travis, the isPlaying test does not always pass due to async issue.

src/base/playback.js
towerz: I agree, there should be a warning message whenever a non-fatal error occurs.

kslimani: I added a Log.warn() in 6e6df23e35c49d67a68bb281790010b4c5dd7bcc

src/base/playback.js
```diff
@@ -210,6 +221,25 @@ export default class Playback extends UIObject {
   destroy() {
     this.remove()
   }
+
+  /**
+   * attempt to autoplays the playback.
+   * @method autoPlay
+   */
+  autoPlay() {
```
towerz: I think we could name this `attemptAutoPlay` to make its purpose clearer.

src/playbacks/html5_video/html5_video.js
```diff
@@ -215,7 +240,10 @@ export default class HTML5Video extends Playback {
   // On mobile device, HTML5 video element "retains" user action consent if
   // load() method is called. See Player.consent().
   consent() {
-    !this.isPlaying() && this.el.load()
+    if (!this.isPlaying()) {
+      this.el.load()
+      super.consent()
```
towerz: I'd just move the `super.consent()` call to before the load call to prevent the `consented` state from being inconsistent during event handling.

kslimani:
I renamed autoplay method and moved consent call. Also made some change to make media reusable. (for example by an external plugin).

kslimani:
I can't figure out why Travis is failing on tests. It success locally. I will try to setup something with chromium 62 to debug this.

kslimani:
@jhonatangomes the test which is mainly failing on Travis is [html5_video isPlaying](https://github.com/clappr/clappr/pull/1644/commits/02f38c6087c9553074bda11042263639b571c759). I tried to fix it, but i am not successfull here.

Locally it passes, but i have Chrome 65 (not 62) on the Linux computer.

jhonatangomes:
I think we can test this way:

``` javascript
it('isPlaying() is true after constructor when autoPlay is true', function(done) {
  const playback = new HTML5Video({ src: 'http://example.com/dash.ogg', autoPlay: true, mute: true })

  setTimeout(function() {
    expect(playback.isPlaying()).to.be.true
    done()
  }, 1000)
})
```

I think this ensures that nextTick from html5_video constructor and play promise will be resolved after 1 second and the test will pass.

kslimani:
In fact, it may be currently better to keep autoplay handle feature decoupled from player, and instead update player documentation providing some code example. (for example using can-autoplay library).

And push this feature in 0.3.x with @towerz work.

I close the PR.

kslimani:
Added an example in [this issue](https://github.com/clappr/clappr/issues/1639#issuecomment-395414240).

towerz:
@kslimani I think we should merge this PR anyway, and perhaps add an option to disable this behavior. It won't interfere with what we'll implement on 0.3.x, and people will benefit from it.

stefanullinger:
Hey guys, good work! One question, though: When `attemptAutoPlay()` and indirectly `canAutoPlay()` are called within the constructor, how do I make sure that [this.consented](https://github.com/clappr/clappr/blob/master/src/playbacks/html5_video/html5_video.js#L161) is true for mobile devices?

Currently, I am trying to call `player.consent()` after initializing the basic player (without specifying any source) and afterwards calling `player.load('someSourceFile', '', true);`.

kslimani:
@stefanullinger on mobile device,  the `player.consent()` must be called as the result of user action (ie: click or touch).

You may be also interested in the `recycleVideo` [playback option](https://github.com/clappr/clappr/blob/master/doc/BUILTIN_PLUGINS.md#playback-configuration).

stefanullinger:
@kslimani I have already set the recycleVideo playback option to true, thanks.

And I am already calling `player.consent()`. But this has no effect for mobile devices as the Playback has not been initialized yet. When I then call `player.load()` the playback get's initialized but does not know about my previous call to `player.consent()`.

kslimani:
@stefanullinger i recommend create a new issue (for better tracking). Also if you could provide an example of what you are trying to achieve.

Reminder: on device mobile, it is not possible to autoplay video on page load (only after user action).

stefanullinger:
@kslimani Okay, I will create a new issue later. [Here](https://jsfiddle.net/stefanullinger/5psv7xu3/embedded/result) is some basic example of what I am trying to do. When you set a breakpoint in `canAutoPlay()` and inspect `this.consented`, you can see it is `false`.

kslimani:
@stefanullinger i see what is going wrong in your example : you create player instance inside click event handler. Player instance (ie: video element) should be created before the user click button.

Please, create a new issue and i will be able to give you more details.

PR #1687, opened 2018-08-10
Added reference to P2P Media Loader GitHub repo

mrlika:
FOSS library that adds P2P stream sharing capabilities of live and VOD HLS streams into Clappr


leandromoreira:
Although I'm very excited with this new FOSS p2p great component, I don't think we can put it under the plugins belt, it's IMO more of a configuration for the most used playbacks (HLSjs and Shaka). @clappr/core what do you think about this? 

I was wondering about putting under the configuration doc or even on the FAQ (as a question)?

mrlika:
I think doc/EXTERNAL_PLUGINS.md is a good place for it. But the library is not a plugin. We can start a new section there and name it "External Integrations" or something similar.

And if people really frequently ask about P2P in Clappr we can add it to FAQ page too.

mrlika:
Created new pull request #1691

PR #1694, opened 2018-08-22
fix: mediacontrol autohide timeout

kslimani:
It adds `hideMediaControlDelay` embed parameter in documentation.

It also fix and issue in the media control render method which does not take account of `hideMediaControlDelay` value. It currently have an hardcoded 1000 milliseconds default value.

__Note:__ i did not change the default value to 2000 (same value as [hide()](https://github.com/clappr/clappr/blob/master/src/components/media_control/media_control.js#L491) method) to avoid breaking change of the default behaviour. But i think it should be also changed to 2000 [instead of 1000](https://github.com/clappr/clappr/blob/master/src/components/media_control/media_control.js#L662).

So if Clappr core think it should be also changed to 2000, tell me and i modify the PR commit.

### DIFF
--- a/doc/BUILTIN_PLUGINS.md
+++ b/doc/BUILTIN_PLUGINS.md
@@ -170,6 +170,8 @@ you might consider subclassing the base `MediaControl` and using your own custom
 
 If you want to disable media control auto hide, add `hideMediaControl: false` in your embed parameters.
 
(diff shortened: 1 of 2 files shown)


jhonatangomes:
I agree that it should be 2000 too, can you update the PR with this change?

kslimani:
Rewrited last commit with timeout increased from 1000 to 2000 milliseconds.

PR #1703, opened 2018-09-21
Fix actualLiveServerTime when no offset supplied

chriswiggins:
When changes were introduced to work with extended attributes in m3u8 playlists, people that weren't using the attribute just saw ":/:" as a time

### DIFF
--- a/src/plugins/seek_time/seek_time.js
+++ b/src/plugins/seek_time/seek_time.js
@@ -84,15 +84,15 @@ export default class SeekTime extends UICorePlugin {
   getSeekTime() {
     let seekTime, secondsSinceMidnight, d, e
     if (this.useActualLiveTime) {
-      if (this.actualLiveServerTimeDiff) {
-        d = new Date(new Date().getTime() - this.actualLiveServerTimeDiff)
-        e = new Date(d)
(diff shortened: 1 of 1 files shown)


src/plugins/seek_time/seek_time.js
```diff
@@ -84,15 +84,15 @@ export default class SeekTime extends UICorePlugin {
...
-        d = new Date(new Date().getTime() - this.actualLiveServerTimeDiff)
-        e = new Date(d)
-        secondsSinceMidnight = (e - d.setHours(0,0,0,0)) / 1000
-      } else if (this.firstFragDateTime) {
+      if (this.firstFragDateTime) {
         e = new Date(this.firstFragDateTime)
         d = new Date(this.firstFragDateTime)
         d.setHours(0,0,0,0)
         secondsSinceMidnight = ((e.getTime() - d.getTime()) / 1000) + this.duration
+      } else {
```
leandromoreira: shouldn't we also check here `if (this.actualLiveServerTimeDiff)` @clappr/core ?

chriswiggins: No, because it is set to 0 on initialisation if it doesn't exist. See line 41

PR #1718, opened 2018-10-18
Add case insenstive check for HLS playback

### LINKED ISSUE #1522: HLS mimetypes case sensitive?
Just hit this issue - for the HTML5 mimetype list, the type for m3u8 is `application/x-mpegurl` but the FlashHLS and HLS canPlay functions both check for exactly `application/x-mpegURL`

If you provide `application/x-mpegurl` as the content type it will fail to return true from canPlay, which shouldn't be the case as it works fine for the HTML5 mimetype checking.

leandromoreira:
It fixes #1522

### DIFF
--- a/src/base/utils.js
+++ b/src/base/utils.js
@@ -241,6 +241,7 @@ export function removeArrayItem(arr, item) {
 
 // find an item regardless of its letter case
 export function listContainsIgnoreCase(item, items) {
+  if (item === undefined || items === undefined) return false
   return items.find((itemEach) => item.toLowerCase() === itemEach.toLowerCase()) !== undefined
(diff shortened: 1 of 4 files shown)


leandromoreira:
the build is passing locally =/ 
screen shot 2018-10-18 at 11 14 11 am

towerz:
@leandromoreira I took the liberty of removing some duplicate commits coming from another PR. The test is still failing, I'm not sure why.
I'll have a better look at this tomorrow, since I cannot reproduce the failure on either my Mac or Windows machines, so I'll have to run them on Linux and see whether I can find out what's going on.

PR #1729, opened 2018-11-13
ci: test Node.js 6, 8, 10 and 11

DanielRuf:
We should test against all current LTS and stable branches of LTS to catch any issues.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,10 @@
 language: node_js
 sudo: false
-node_js: "8"
+node_js:
+  - "node"
+  - "10"
+  - "8"
+  - "6"
 cache: yarn
 
 before_install:


DanielRuf:
> error upath@1.0.4: The engine "node" is incompatible with this module. Expected version ">=4 <=9".

Well, seems we have to update the dependencies.

DanielRuf:
@towerz these are the current LTS and stable branches of Node.js. Also see https://github.com/clappr/clappr/pull/1729#issuecomment-438226423

It's ,eant to catch issues that exist there as we (contributors and devs and also CI) often require / import the source files for threeshaking and bundling, Node.js projects and different CI platforms.

As you can see users Node.js 10 and newer will have problems.

CI is for testing (also) the different (CI) platforms.

DanielRuf:
For the deploy / publish task, you can limit this to only one of them.

towerz:
@DanielRuf I know what you mean, but I'm opposed to the overhead of testing against ALL of the different LTS versions against the benefit of having those tests run. What I believe is that doing that you're just testing how bundling behaves using those versions, so I'm not actually validating whether the code works, but how the tools perform using a different setup.

Since we just use node 8.x to integrate and build the player itself, the default CI environment should replicate this. If someone is integrating using another version of node.js, it is their responsibility to run tests using this setup.

I'm not sure the overhead is worth it.

DanielRuf:
> Since we just use node 8.x

Shouldn't we switch to 10? It's probably better to use the latest LTS.

towerz:
@DanielRuf I was already running some tests in order to switch to 10. We'll update soon.

DanielRuf:
Ok, then let's close this PR here. Do you plan to constrain the version also using the `engines` field of package.json?

towerz:
@DanielRuf we were considering doing that, yes.

PR #1748, opened 2018-12-20
fix(playback) HTML5 playback isMuted function logic is inverted

bikegriffith:
It looks like it's been (broken) this way for a long time.

In addition, I observed that the `video.muted` property could be set to true while the `video.volume` > 0, and it still behaves muted.  We should check both of the properties.


towerz:
Thanks, @bikegriffith! 🍻

PR #1766, opened 2019-02-13
browser.js -> getBrowserData: replaced for..in by for..of

lmancel:
for..in loop would sometimes retrieve enumerable properties that do not exist on the BROWSER_DATA Array


lmancel:
Linked to issue https://github.com/clappr/clappr/issues/1767

towerz:
Thanks, @lmancel! 🍻

PR #1778, opened 2019-03-26
Listen for CORE_ACTIVE_CONTAINER_CHANGED on Core

chooh:
Event `CORE_ACTIVE_CONTAINER_CHANGED` is triggered on the `Core` object. That's why it's needed to listen for this event of the `Core` object also :-)

### DIFF
--- a/src/components/player.js
+++ b/src/components/player.js
@@ -295,7 +295,7 @@ export default class Player extends BaseObject {
     else
       this._onReady()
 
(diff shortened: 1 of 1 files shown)


chooh:
I tried to create something like a playlist. So I listen for `PLAYER_ENDED`, make an AJAX request and do `TVApp.player.load` with a new stream url.

Then I found than `PLAYER_ENDED` fired only the first time and never after `load()` call. I dug through the source code and found that event `CORE_ACTIVE_CONTAINER_CHANGED` is firing perfectly, but it didn't run `_containerChanged` → `_addEventListeners` → `_addContainerEventListeners`.

Issue #1781, opened 2019-04-11
Double click to go Fullscreen pause/play


BraveSauce:
**Browser**: any

**OS**: any

**Clappr Version**: 0.2.25

**Steps to reproduce**:

* start playing a video
* double click (or double tap) on the container
* it should just go fullscreen without affecting playback, but it pauses and resumes instead

Did you try to reproduce this issue at http://cdn.clappr.io/
-Yep

-------------------------

It's not a real issue on archived content, but when watching a live stream it loses sync and one has to click "back to live" every time the player goes in and out of fullscreen mode.

Is there a way to change the behavior of "pause" and delay it just slightly, like 200ms in order to work around the issue? I tried calling player.play() and then player.pause() with a timeout, on pause event, but it appears that player.play() just overrides it and won't let it pause afterwards.

Any other way to do this?

kslimani:
A quick solution could be to disable the "click to pause" internal plugin.

Try this code on [demo page](clappr.io/demo/) :

```javascript
var playerElement = document.getElementById("player-wrapper");

var player = new Clappr.Player({
  source: 'http://clappr.io/highline.mp4',
  height: 360,
  width: 640,
  events: {
    onReady: function() {
      var p = this.getPlugin('click_to_pause');
      p && p.disable();
    }
  },
});

player.attachTo(playerElement);
```

BraveSauce:
I did that, but then I realized it breaks functionality on mobile, since there is a bug when tapping on the play/pause button which wont let the media control bar hide until you change fullscreen state. So being able to tap on the video to play or pause is still essential in order for the player to work properly. I might open another issue for this, but for the moment I would just like to know whether I can delay the pause event.

kslimani:
A solution could be to override the "click to pause" plugin by a custom one which satisfy your need.

You could also make a PR which modify the existing, and eventually adding a new player option to configure the delay (which default to zero to keep current behaviour by default).

BraveSauce:
> You could also make a PR which modify the existing, and eventually adding a new player option to configure the delay (which default to zero to keep current behaviour by default).

Could you please elaborate on this? (I'm new to clappr)

kslimani:
In Clappr player you can override internal plugin with a third party one if the name is the same.

So if you create your own playback plugin and name it "click_to_pause" it will replace the internal one.

[...]

PR #1784, opened 2019-04-16
fix(container): handle double touch event

kslimani:
Since the touch module [has been removed](https://github.com/clappr/clappr-zepto/commit/7dca838af14dab84ebbadef79a27abf45b70553d) from Clappr Zepto custom build, the container double tap/touch event is broken.

I googled about double tab and i make a quick fix based on this [jsFiddle](http://jsfiddle.net/brettwp/J4djY/).

I am not sure if this is the most elegant way to fix the issue, but it seems to work _(i tested on samsung galaxy, galaxy tab, iphone and ipad)_.

Should satisfy #1781.

### DIFF
--- a/src/base/utils.js
+++ b/src/base/utils.js
@@ -316,22 +316,16 @@ export class DoubleEventHandler {
   constructor(delay = 500) {
     this.delay = delay
     this.lastTime = 0
-    this.timer = null
   }
 
   handle(event, cb, prevented = true) {
     // Based on http://jsfiddle.net/brettwp/J4djY/
     let currentTime = new Date().getTime()
     let diffTime = currentTime - this.lastTime
-    clearTimeout(this.timer)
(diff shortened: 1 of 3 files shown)


kslimani:
I made a small behaviour change in container component : the container click event is cancelled by the double click/tap event.

This change prevents playback from pausing when switching to full screen. _(should improve user experience)_

I tested on many devices, but it would be great if someone else could also perform tests and give some feedbacks.

kslimani:
Moved double touch event handling logic into base utils module _(to allow re-use)_

towerz:
Thanks, @kslimani! 🍻 

This is a great solution. We're in the process of redesigning Clappr render/interaction model to work better with some render frameworks. Probably we'll be able to reuse this module in the DOM-based model.

PR #1800, opened 2019-06-01
HLS - Fix interval timer leak

jhonatangomes:
This PR fixes a leak of the timer from the HLS plugin. 

The problem happens when the method 'play` is called multiple times. This scenario can happen when the user pauses/play the video, for example.

For each time the `play` was being called, a new timer was being created and not being cleared when the playback was being destroyed.

This PR ensures that only one timer is running and, also, stop the timer when the playback is stopped.


towerz:
Nice catch, @jhonatangomes! 🍻

Issue #1809, opened 2019-06-13
HLS errors


ctaity:
**Browser**: ALL

**OS**: ALL

**Clappr Version**: lastest

**Steps to reproduce**:
have a m3u8 link return 403

* I was expecting player call Clappr.Events.PLAYBACK_ERROR.

Hi, we have m3u8 with aes128 encription and we deny keys based on ips, browser etc. When we deny a key, clappr logs: 
  clappr.min.js:1 [warn][hlsjs: non-fatal error occurred, evt hlsError, data [object Object] ]
  clappr.min.js:1 [debug][hls] playback:levels:available (5) [{…}, {…}, {…}, {…}, {…}]

Take this error as non fatal error, but without key can play de the video. I think in this case the error is fatal.

kslimani:
When i load a m3u8 returning a 403 forbidden i got a PLAYBACK_ERROR.

Using this code on [Clappr demo page](http://clappr.io/demo/) :

```javascript
var playerElement = document.getElementById("player-wrapper");

var player = new Clappr.Player({
  source: 'http://rbetis-hls.flumotion.com/rbetis/live/.m3u8',
  height: 360,
  width: 640,
  events: {
    onReady: function() {
      this.core.activePlayback.listenTo(
        this.core.activePlayback,
        Clappr.Events.PLAYBACK_ERROR,
        function(e) {
          console.log('playback_error', e);
        }
      );
    },
  },
});

player.attachTo(playerElement);
```

ctaity:
Thanks for the reponse, but the error is on the aes 128 key request, we deny the key based on the ip for geolocation. 
 
On hls.js on the function   _onHLSJSError(evt, data), exits on: 
```
} else {
      error.level = PlayerError.Levels.WARN
      this.createError(error)
      Log.warn('hlsjs: non-fatal error occurred', { evt, data })
    }
```

I think, almost can trigger  playback error event with non-fatal to trap or something like thath. 
The error from HLS.JS lib is 

```javascript
  if (data.type == "networkError" && data.details == "keyLoadError") 
```

kslimani:
Oh, i see.

I agree, denying the key in encrypted HLS source scenario could also trigger a playback fatal error. (it is technically near like returning 403 on manifest).

If you can share an url which reproduce the issue i can propose a fix. _(privately via email)_

ctaity:
[...]

PR #1811, opened 2019-06-18
feat: add triggerFatalErrorOnResourceDenied playback option

kslimani:
On iOS devices, HLS source triggers a fatal playback error if encryption key is denied.

This new `triggerFatalErrorOnResourceDenied` playback option, when set to true, will attempt to replicate this behaviour (ie: trigger playback fatal error).

It should satisfy #1809.

### DIFF
--- a/src/playbacks/hls/hls.js
+++ b/src/playbacks/hls/hls.js
@@ -335,10 +335,32 @@ export default class HLS extends HTML5VideoPlayback {
         this.stop()
       }
     } else {
+      // Transforms HLSJS.ErrorDetails.KEY_LOAD_ERROR non-fatal error to
+      // playback fatal error if triggerFatalErrorOnResourceDenied playback
+      // option is set. HLSJS.ErrorTypes.KEY_SYSTEM_ERROR are fatal errors
+      // and therefore already handled.
+      if (this.options.playback
+        && this.options.playback.triggerFatalErrorOnResourceDenied
+        && this._keyIsDenied(data)
+      ) {
+        Log.error('hlsjs: could not load decrypt key.', { evt, data })
+        formattedError = this.createError(error)
+        this.trigger(Events.PLAYBACK_ERROR, formattedError)
+        this.stop()
+        return
+      }
+
       error.level = PlayerError.Levels.WARN
       this.createError(error)
       Log.warn('hlsjs: non-fatal error occurred', { evt, data })
     }
+  }
+
+  _keyIsDenied(data) {
+    return data.type === HLSJS.ErrorTypes.NETWORK_ERROR
+      && data.details === HLSJS.ErrorDetails.KEY_LOAD_ERROR
+      && data.response
+      && data.response.code >= 400
   }
 
   _onTimeUpdate() {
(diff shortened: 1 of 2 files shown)


kslimani:
Do not merge : forgot to set error message string.

src/playbacks/hls/hls.js
```diff
@@ -335,12 +335,34 @@ export default class HLS extends HTML5VideoPlayback {
         this.stop()
       }
     } else {
+      // Transforms HLSJS.ErrorDetails.KEY_LOAD_ERROR non-fatal error to
+      // playback fatal error if triggerFatalErrorOnResourceDenied playback
+      // option is set. HLSJS.ErrorTypes.KEY_SYSTEM_ERROR are fatal errors
+      // and therefore already handled.
+      if (this.options.playback
+        && this.options.playback.triggerFatalErrorOnResourceDenied
```
leandromoreira: Shouldn't we explicity set this to false as default at some init code? do we do already that @clappr/core for other options?

kslimani: It seems that not every options are initialized. Sometimes it uses a getter. This new option is supposed to be a bool value, it's the main reason it is not initialized. I you prefer have it initialized to false in constructor, just tell me and i add the fix.

leandromoreira: I agree with you =) let's see if the other think the same way and maybe we hack this `feature` into this PR lol

towerz: We usually don't initialize those nested options to keep initialization simple. What we could do is to always have `this.options.playback` initialized to avoid having to check whether it is defined whenever we want to access it.

@kslimani @leandromoreira, what do you think?

leandromoreira: Maybe we could have this notion of `option` as first class citizen where we could initialize the defaults:

```javscript

class Option
   const defaults = {
      option: 1,
      hls: {
         option1: 1
      }
   }
   function init()
end

```

It'd make easy to document/read and to centralize any kind of option's behavior.

leandromoreira: Or maybe we should create a PR for this lol

leandromoreira: The `player` component is nice place to see how nicely it works for documentation https://github.com/clappr/clappr/blob/dev/src/components/player.js#L153

towerz: @kslimani I'd avoid creating a first-class citizen for the moment since we'll probably split the options that are visible for each component. We still need to think whether we'll need another component responsible for that or what components will hold references to the options set.

kslimani: @leandromoreira @towerz initializing _[scope]_.options.playback to empty object if undefined is a good idea. Assuming this object is defined allow to simplify many conditions. I too think this initialization could be done in player component, or delegated to core, but in another PR. _(this new PR could introduce the initialization and update other classes conditions where applicable)_

kslimani: In fact, it seems playback option object is [already initialized](https://github.com/clappr/clappr/blob/dev/src/components/player.js#L240).

towerz: @kslimani indeed it is... we made that to merge compatible options that were added without the `playback` scope. We intend to remove it for `0.4.x` where we'll have a bunch of breaking changes in a few weeks.

towerz:
Thanks, @kslimani! 🍻

PR #1815, opened 2019-06-24
docs(builtin-plugin): set recycleVideo default value to true

### LINKED ISSUE #1813: Video element recycling enabled by default
While browsing code about default option initialization, i found that `recycleVideo` playback option is initialized to `true` by default, since [this commit](https://github.com/clappr/clappr/commit/9d2d307ff7c1c2a90f5d5afa9bd041c6eef2ac6c#diff-200a325d8607daea7e681ef9d1ab1426).

In the past, i set this option to `false` by default for backward compatibility reason, but also because i found bugs with some browsers like Firefox (desktop). I recommend to enable this option only with mobile/tablet devices.

@towerz @leandromoreira @joaopaulovieira : could you confirm this option should still be enabled by default ? _(if yes, i will make a PR to update documentation)_.

kslimani:
Update documentation.

Closes #1813.

### DIFF
--- a/doc/BUILTIN_PLUGINS.md
+++ b/doc/BUILTIN_PLUGINS.md
@@ -81,7 +81,7 @@ The configuration for the playback, it's still only compatible with `html5_video
     controls: true,
     playInline: true, // allows inline playback when running on iOS UIWebview
     crossOrigin: 'use-credentials',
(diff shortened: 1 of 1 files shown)


towerz:
Thanks, @kslimani! 🍻

PR #1830, opened 2019-07-26
fix(poster): avoid conflict with click_to_pause plugin

### LINKED ISSUE #1829: bug(poster-plugin): click event conflicts with click_to_pause if audio only
**Browser**: any

**OS**: any

**Clappr Version**: latest

**Steps to reproduce**:

* load an audio only source, for example,  `https://file-examples.com/wp-content/uploads/2017/11/file_example_MP3_5MG.mp3`

Video will start when clicking on poster, but then since poster is [still displayed with audio](https://github.com/clappr/clappr/blob/dev/src/plugins/poster/poster.js#L108), the click event will not be propagated to `click_to_pause` plugin. _(with video source, the poster css display value is none)_.

kslimani:
Should fix #1829 issue.

### DIFF
--- a/src/plugins/poster/poster.js
+++ b/src/plugins/poster/poster.js
@@ -95,12 +95,15 @@ export default class PosterPlugin extends UIContainerPlugin {
   }
 
   clicked() {
-    if (!this.options.chromeless || this.options.allowUserInteraction) {
-      this.playRequested = true
-      this.update()
-      this.container.play()
+    // Let "click_to_pause" plugin handle click event if media has started playing
+    if (! this.hasStartedPlaying) {
+      if (!this.options.chromeless || this.options.allowUserInteraction) {
+        this.playRequested = true
+        this.update()
+        this.container.play()
+      }
+      return false
     }
-    return false
   }
 
   shouldHideOnPlay() {


kslimani:
Travis seems to have failed because `/etc/init.d/xvfb` may be missing.

src/plugins/poster/poster.js
```diff
@@ -95,12 +95,15 @@ export default class PosterPlugin extends UIContainerPlugin {
   }
 
   clicked() {
-    if (!this.options.chromeless || this.options.allowUserInteraction) {
-      this.playRequested = true
-      this.update()
-      this.container.play()
+    // Let "click_to_pause" plugin handle click event if media has started playing
+    if (! this.hasStartedPlaying) {
```
leandromoreira: can't we inline this condition?
```js
if (! this.hasStartedPlaying && (!this.options.chromeless || this.options.allowUserInteraction)) {}
```

src/plugins/poster/poster.js
```diff
@@ -95,12 +95,15 @@ export default class PosterPlugin extends UIContainerPlugin {
...
+      if (!this.options.chromeless || this.options.allowUserInteraction) {
+        this.playRequested = true
+        this.update()
+        this.container.play()
+      }
+      return false
     }
```
leandromoreira: Shouldn't we always return something? I think the last `return` is missing

kslimani: @leandromoreira if you return false, then click event is not propagated  (and therefore click_to_pause plugin click is not triggered).

src/plugins/poster/poster.js
kslimani: The condition could be rewrited, maybe removing the return false because of hasStartedPlaying check. I writed it this way to ensure it do not break anything.

kslimani: Also if you rewrite the condition that way, with audio only, if you seek almost to end, then wai for video to stop and display poster play icon, then if you click, it does not start the video.

kslimani: @leandromoreira  In fact, the return false is required when click is handled by poster. Returning false disrupt the click event and container click if not triggered. Otherwise, if triggered the media is immediatly paused by `click_to_pause` plugin.

Just experiment the branch by yourself :)

towerz:
@kslimani our travis config has been updated to fix an issue with PR checks. Could you update your branch with the latest `dev`?

Issue #1243, opened 2016-12-26
When having two player instances playing, stopping one instance triggers exiting fullscreen mode on both


Eetin:
**Browser**: 54.0.2840.100 (64-bit)

**OS**: Ubuntu 16.04 LTS

**Clappr Version**: 0.2.64

**Steps to reproduce**:

* Create two player instances and start both
* Switch first player to fullscreen mode
* Stop the second player
* I was expecting the first player to remain in fullscreen mode but it doesn't

kslimani:
Currently, [Clappr player](https://github.com/clappr/clappr/blob/master/src/base/utils.js#L89) rely on prefixed _fullscreenElement_ [document api attributes](https://developer.mozilla.org/en-US/docs/Web/API/Fullscreen_API#Prefixing) to detect if browser is in fullscreen : _(webkitFullscreenElement, mozFullScreenElement, msFullscreenElement, webkitFullscreenElement, webkitFullscreenElement)_ - Which explains the current "global" behaviour.

If attribute is null, then browser is not in fullscreen mode, otherwise the attribute contains the element which has requested fullscreen. 

@core : To fix/implement this behaviour, maybe add a unique id _(a data-XXX attribute ?)_ to playback element and check for it if attribute is not null ? _(not sure it is the best/elegant way to do)_

leandromoreira:
@kslimani you're right about the reason, good catch.

kslimani:
If i am not wrong, enter fullscreen require a user action (ie: click or touch). So if one player is in fullscreen, the scenario you describe cannot happen. (you can't click on the other player if other one is in fullscreen)

luanraithz:
I'm sorry, I thought that there could be cases where you call the method to enter full screen.

kslimani:
You can request fullscreen programmatically but it will be denied by the browser if it is not trigger from user click.

Another possible funny scenario could be to create a plugin which add a new button in first player media controls which trigger fullscreen on second player :smile:

luanraithz:
Ok I think I open an PR this week about this issue, I just had do make sure what was supposed to be the behavior.

PR #1863, opened 2019-10-03
Fixed problem with multiple instances of clappr and fullscreen mode

luanraithz:
As mentioned in #1243, when having two instances, once one stops the other one leaves fullscreen mode.
Can be reproduced inside the demo:

Pasting and running the following code in the editor area:
```javascript
var playerElement1 = document.createElement(`div`)

playerElement1.id = "player-wrapper1"
document.body.appendChild(playerElement1)

var player1 = new Clappr.Player({
  mute: true,
  height: 360,
  width: 640
});

player1.attachTo(playerElement1);
```
Start `player` , wait some seconds, start the `player1`, enter fullscreen mode with the `player1` and wait until the `player` has ended.

The fix was just validating if the current instance container is the one that is fullscreen, any better solution would be welcome!


test/components/core_spec.js
```diff
@@ -75,11 +75,13 @@ describe('Core', function() {
...
         it('adds a class "fullscreen" to core element', () => {
           const spy = sinon.spy(this.core.$el, 'addClass')
+          sinon.stub(this.core, 'isFullscreen').returns(false)
+
           expect(spy).not.to.have.been.called
 
           this.core.toggleFullscreen()
-
           expect(spy).to.have.been.calledWith('fullscreen')
+
```
towerz: There are a number of those empty lines at the end of blocks, please remove those.

test/components/core_spec.js
```diff
@@ -95,22 +97,25 @@ describe('Core', function() {
           expect(fullScreenSpy).to.have.been.calledWith(this.core.activeContainer.el)
         })
       })
+
```
towerz: There are a number of those empty lines at the end of blocks, please remove those.

test/components/core_spec.js
```diff
@@ -95,22 +97,25 @@ describe('Core', function() {
...
-      })
 
       it('calls Fullscreen.cancelFullscreen', () => {
         const spy = sinon.spy(Fullscreen, 'cancelFullscreen')
+        sinon.stub(Browser, 'isiOS').value(false)
+        sinon.stub(this.core, 'isFullscreen').returns(true)
+
         this.core.toggleFullscreen()
         expect(spy).to.have.been.called
+
```
towerz: There are a number of those empty lines at the end of blocks, please remove those.

test/components/core_spec.js
```diff
@@ -119,6 +124,23 @@ describe('Core', function() {
...
+        fakeContainer1.setAttribute('id', 'fakeContainer1')
+        newInstance.el = fakeContainer1
+
+        expect(this.core.isFullscreen()).to.equal(false)
+        expect(newInstance.isFullscreen()).to.equal(false)
+
+        sinon.stub(Fullscreen, 'getFullscreenElement').returns(fakeContainer1)
+        expect(this.core.isFullscreen()).to.equal(false)
+        expect(newInstance.isFullscreen()).to.equal(true)
+
```
towerz: There are a number of those empty lines at the end of blocks, please remove those.

rodrigobacelli:
@towerz .editorconfig file in the project enforces this space at the end of file. It would be better if `insert_final_newline` is set as false, so IDEs will not add this line.

towerz:
@rodrigobacelli I don't think you understood. I was complaining about empty lines **at the end of blocks**, not files.

rodrigobacelli:
> @rodrigobacelli I don't think you understood. I was complaining about empty lines **at the end of blocks**, not files.

Yeah, you're completely right. Sorry, my bad

src/components/core/core.js
```diff
@@ -308,13 +308,17 @@ export default class Core extends UIObject {
...
+  isFullscreen() {
+    return Fullscreen.getFullscreenElement() === (Browser.isiOS ? this.activeContainer.el : this.el)
+  }
+
   toggleFullscreen() {
-    if (!Fullscreen.isFullscreen()) {
-      Fullscreen.requestFullscreen(Browser.isiOS ? this.activeContainer.el : this.el)
-      !Browser.isiOS && this.$el.addClass('fullscreen')
-    } else {
+    if (this.isFullscreen()) {
```
luanraithz: @towerz Do you think that this validation should occur in both places (plugin and core)?
The plugin doesn't actually need to `toggleFullscreen`, it always want to cancel it.

towerz: @luanraithz the validation should be in both because they aren't due to the same reason. In the Core `toggleFullscreen` the check is necessary to determine whether to enter or exit fullscreen.

In the case of the plugin, the check is necessary to avoid an unnecessary call to `cancelFullscreen`.

test/components/core_spec.js
```diff
@@ -119,6 +121,22 @@ describe('Core', function() {
           expect(spy).to.have.been.calledWith('fullscreen nocursor')
         })
       })
+
```
towerz:
```suggestion
```
Please remove this extra empty line.

src/base/utils.js
```diff
@@ -63,6 +63,12 @@ export const Fullscreen = {
       document.msFullscreenElement
     )
   },
+  getFullscreenElement: function() {
```
towerz: Since you're adding the `getFullscreenElement` helper, the `isFullscreen` function should use it to avoid repeated code.

towerz:
Thanks, @luanraithz! 🍻

PR #1871, opened 2019-10-20
Added code formatter with a pre-commit tool

marcelometal:
Added prettier + husky + lint-staged


towerz:
@marcelometal thanks for the contribution, but we're still reviewing some of our code styles, and we don't think using prettier is the direction we want to follow. We have a specific code style we're trying to implement and using an opinionated tool won't help us with that. For example, we want to maximize visualization for wider screen sizes with longer line lengths.

There are some other details, like enforcing formatting on commit time that could generate larger diffs than what we want, like your own commit, with 86 modified files. This messes a lot with the file history, and we don't believe this large changeset will be helpful.

Last, but not least, we're splitting the project codebase into separate repositories, mostly focused on a given feature set. The new repositories will share code style and other settings, but most of the code from this repository will be moved away. So, it doesn't make much sense to accept this PR right now.

Issue #1846, opened 2019-08-31
UI customization


DanieleSuppo73:
I'm sorry to write this question on issues, but I can't find any other place to do that.
How can I customize the play button and any other UI elements?
Many thanks!

towerz:
Hi, @DanieleSuppo73. Right now, we don't have the UI elements separate from the `MediaControl` plugin. So the way to do that right now is extending it or replacing with your own controls.

Extending would be the easiest way because you would be able to reuse the template and only replace the UI elements or CSS styles as needed.

In a future release (probably this month) we'll have the UI elements as separate plugins, so you would only need to override the elements you want to replace or change instead of overriding the entire UI.

ChristopheBougere:
I'm just discovering clappr and I have the same need, I can't wait for the new release that makes it easy to customize UI 👌

kslimani:
Maybe make some non-breaking changes, for example expose all svg icons in a class or a simple plain object in the Utils module (Utils.SVG ?) and use Utils module in media_control, favicon, poster, closed_captions and arror_screen plugins.

This simple change could allow developper to override svg with customs. For example :

```javascript
Clappr.Utils.SVG.play = 'raw_custom_svg_import'; // before create Clappr instance
```

It is pretty simple changes, i can submit a PR proposition if required.

PR #1877, opened 2019-11-04
feat(utils): add SVG icons module

kslimani:
A simple proposition which should satisfy #1846.

It's a simple workaround to allow developper to override player SVG icons. _(in the meantime to implement a proper way)_.

Example usage :

```javascript
var customPlay = '<svg version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" x="0px" y="0px" viewBox="0 0 492.719 492.719" xml:space="preserve"><path d="M492.719,166.008c0-73.486-59.573-133.056-133.059-133.056c-47.985,0-89.891,25.484-113.302,63.569 c-23.408-38.085-65.332-63.569-113.316-63.569C59.556,32.952,0,92.522,0,166.008c0,40.009,17.729,75.803,45.671,100.178 l188.545,188.553c3.22,3.22,7.587,5.029,12.142,5.029c4.555,0,8.922-1.809,12.142-5.029l188.545-188.553 C474.988,241.811,492.719,206.017,492.719,166.008z"/></svg>';

// Override before create Clappr instance
Clappr.Utils.SvgIcons.play = customPlay;

// A small heart will replace the default play icon
  // [...]
});
```

### DIFF
--- a/src/base/svg_icons.js
+++ b/src/base/svg_icons.js
@@ -1,0 +1,38 @@
+// Copyright 2014 Globo.com Player authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+import playIcon from '../icons/01-play.svg'
+import pauseIcon from '../icons/02-pause.svg'
+import stopIcon from '../icons/03-stop.svg'
+import volumeIcon from '../icons/04-volume.svg'
+import volumeMuteIcon from '../icons/05-mute.svg'
+import fullscreenIcon from '../icons/06-expand.svg'
+import exitFullscreenIcon from '../icons/07-shrink.svg'
+import hdIcon from '../icons/08-hd.svg'
+import ccIcon from '../icons/09-cc.svg'
+import reloadIcon from '../icons/10-reload.svg'
+
+export let cc = ccIcon
+export let exitFullscreen = exitFullscreenIcon
+export let fullscreen = fullscreenIcon
+export let hd = hdIcon
+export let pause = pauseIcon
+export let play = playIcon
+export let reload = reloadIcon
+export let stop = stopIcon
+export let volume = volumeIcon
+export let volumeMute = volumeMuteIcon
+
+export default {
+  cc: ccIcon,
+  exitFullscreen: exitFullscreenIcon,
+  fullscreen: fullscreenIcon,
+  hd: hdIcon,
+  pause: pauseIcon,
+  play: playIcon,
+  reload: reloadIcon,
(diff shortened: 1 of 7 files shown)


kslimani:
The `Clappr.Utils.SvgIcons` properties are :

```javascript
{
  cc: 'the closed captions icon',
  exitFullscreen: 'the exit fullscreen icon',
  fullscreen: 'the fullscreen ison',
  hd: 'the HD icon',
  pause: 'the pause icon',
  play: 'the play icon',
  reload: 'the reload icon',
  stop: 'the stop icon',
  volume: 'the volume icon',
  volumeMute: 'the volume mute icon',
}
```

kslimani:
@towerz yes you already mentionned about spliting. This PR is just a simple temporary workaround.

I partially understand the requirement to provide customization at plugin level to keep only icons to load. What about plugin which share the same icon ? For exemple media_controls and poster plugin currently share the same play icon; if plugins are splited, how to you plan to bundle the icons (will they be duplicate in each plugin ?)

towerz:
@kslimani if it's a temporary workaround, I think we can add it. But with a notice that it will only work in the current version since it would be removed on `0.4.x`, which I was just holding until PRs are merged and a final `0.3.x` release is made.

About the icons that are shared between plugins, they will be shared modules. In the current form of the clappr plugins project, they will be bundled together and deduplicated, but when the plugins are split apart, the icon may be duplicated in each package, or added to a separate package instead. I prefer the second approach to prevent duplicated code in the final bundle.

kslimani:
@towerz i confirm that this PR was just an "immediate" workaround to the #1846 issue _(ie: for example simply change the play icon, which seems to be a common need)_.

Indeed theses change will be removed in 0.4.x (with probably more breaking changes ?)

If you think that 0.4.x will be released soon, then feel free to close it, otherwise i think it will be usefull in the meantime. :+1:

towerz:
@kslimani I think we can go with this solution since it's very simple, and there are some use cases that would benefit from this already.

Since `0.4.x` will introduce a number of breaking changes we can include this one in the list as well. I expect to release it by the end of the week, but I'll still have some work to isolate the icons and provide a similar way to solve this issue.

So, let's do this! 🍻

Hrishikesh-K:
@towerz @kslimani sorry to bother you if it's already covered somewhere, but, I couldn't find it. If Clappr 0.4.2 does not support the way mentioned here, how exactly is it to be done in the latest version? Thanks for the help.

PR #1879, opened 2019-11-06
fix: repair fullscreen element detection with Firefox

kslimani:
If fix fullscreen detection with Firefox browser. _(based on current fullscreen element)_

It also rename the Utils.Fullscreen `getFullscreenElement` method to `fullscreenElement` to be more consistent with [fullscreen api](https://developer.mozilla.org/en-US/docs/Web/API/DocumentOrShadowRoot/fullscreenElement).

### DIFF
--- a/src/base/utils.js
+++ b/src/base/utils.js
@@ -56,10 +56,10 @@ export function formatTime(time, paddedHours) {
 }
 
 export const Fullscreen = {
-  getFullscreenElement: function() {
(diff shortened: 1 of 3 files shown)


towerz:
Nice catch! Thanks, @kslimani :beers:

andrepolischuk:
hi, are you planning to release this soon? thanks 😁

PR #1880, opened 2019-11-08
Fix: sync stored volume value with the value passed to the playback and allow floats

RussCoder:
Current behavior:

```js
player.setVolume(0.9); // the volume will be set, you will hear the sound, but ...
player.getVolume(); // will return 0 
// and after:
player.mute();
player.unmute(); 
// you don't hear the sound anymore, since the stored value 0 is really set as the volume now
 ```

This PR fixes it and allows to use floats normally. To my mind, 1 is much louder than 0.5, so a float volume is a useful feature.

### DIFF
--- a/src/components/container/container.js
+++ b/src/components/container/container.js
@@ -388,9 +388,9 @@ export default class Container extends UIObject {
   }
 
   setVolume(value) {
-    this.volume = parseInt(value, 10)
-    this.trigger(Events.CONTAINER_VOLUME, value, this.name)
-    this.playback.volume(value)
+    this.volume = parseFloat(value)
+    this.trigger(Events.CONTAINER_VOLUME, this.volume, this.name)
+    this.playback.volume(this.volume)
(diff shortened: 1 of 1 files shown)


kslimani:
@RussCoder i think the expected argument value is a percentage. (ie: an integer between 0 and 100).

RussCoder:
@kslimani maybe. But now de facto floats are allowed: you can set a float value. And if you don't mute/unmute a player, you perhaps, may never even know that floats are not implied as valid values. 

If you now change the behavior - sync the stored value and the value passed the playback, but make it always integer, it may be a kind of breaking change for those who use values like 0.5 now, since such values works properly.

And with my fix, the current behavior remains untouched - you still can use only integers. And floats work still. But stored value corresponds to the real volume level and mute/unmute don't change the real volume level.

towerz:
Thanks, @RussCoder :beers:

PR #1918, opened 2020-04-11
fix: Fullscreen error on double click

### LINKED ISSUE #1860: Fullscreen error on double click
**Browser**: Chrome 77.0.3865.90 

**OS**: Windows 10

**Clappr Version**: 0.3.9

**Steps to reproduce**:

* first step
Open console then double click on player:

Uncaught (in promise) TypeError: fullscreen error

Did you try to reproduce this issue at http://cdn.clappr.io/

jamalosm:
Fixes #1860


jamalosm:
The "bindContainerEvents" method was called 2 times, so events were listened twice

luzeduardo:
@ Hi @pedrochamberlain . Could you check it again? I've tested on Chrome `81.0.4044.138` and Firefox `76.0.1` successfully.

pedrochamberlain:
> @ Hi @pedrochamberlain . Could you check it again? I've tested on Chrome `81.0.4044.138` and Firefox `76.0.1` successfully.

Tested the fix again and found that it works, must've scrambled branches during testing, mistake on my part.

pedrofurtado:
Any news about this PR? Happy to see this merged and released :tada:

luanraithz:
As far as I know the content inside the `dist` folder shouldn't be pushed, I think it should be changed on releases.
@towerz Can you confirm that?

pedrofurtado:
@leandromoreira @towerz Is this PR ok for merge? 👍

kslimani:
Looking at this PR, i think this PR should not include dist folder files.

Also, i hardly understand how removing the binding of container events in the mediacontrols module fix the issue 🤷 _(and it may introduce more issues ?)_

According #1860 , it seems to be an error throw by some Promise when switching to fullscreen.

leandromoreira:
> As far as I know the content inside the dist folder shouldn't be pushed, I think it should be changed on releases.

can you remove the dist? @pedrochamberlain

kslimani:
IMO, a more proper way to fix this issue would be to simply catch the [requestFullScreen](https://developer.mozilla.org/en-US/docs/Web/API/Element/requestFullScreen) Promise error in the base [utils](https://github.com/clappr/clappr/blob/dev/src/base/utils.js#L67) module.

For example :

```diff
diff --git a/src/base/utils.js b/src/base/utils.js
index bb49a69..385f451 100644
--- a/src/base/utils.js
+++ b/src/base/utils.js
@@ -63,9 +63,10 @@ export const Fullscreen = {
       document.msFullscreenElement
   },
   requestFullscreen: function(el) {
-    if (el.requestFullscreen)
-      el.requestFullscreen()
-    else if (el.webkitRequestFullscreen)
+    if (el.requestFullscreen) {
+      let p = el.requestFullscreen()
+      p.catch && (p.catch(e => { /* Fixes #1860 */ }))
+    } else if (el.webkitRequestFullscreen)
       el.webkitRequestFullscreen()
     else if (el.mozRequestFullScreen)
       el.mozRequestFullScreen()
```

leandromoreira:
> IMO, a more proper way to fix this issue would be to simply catch the [requestFullScreen](https://developer.mozilla.org/en-US/docs/Web/API/Element/requestFullScreen) Promise error in the base [utils](https://github.com/clappr/clappr/blob/dev/src/base/utils.js#L67) module.
> 
> For example :
> 
> ```diff
> diff --git a/src/base/utils.js b/src/base/utils.js
> index bb49a69..b94eb9f 100644
> --- a/src/base/utils.js
> +++ b/src/base/utils.js
> @@ -63,9 +63,10 @@ export const Fullscreen = {
>        document.msFullscreenElement
>    },
>    requestFullscreen: function(el) {
> -    if (el.requestFullscreen)
> -      el.requestFullscreen()
> -    else if (el.webkitRequestFullscreen)
> +    if (el.requestFullscreen) {
> +      let p = el.requestFullscreen()
> +      p && p.catch && (p.catch(e => { /* Fixes #1860 */ }))
> +    } else if (el.webkitRequestFullscreen)
>        el.webkitRequestFullscreen()
>      else if (el.mozRequestFullScreen)
>        el.mozRequestFullScreen()
> ```

thoughts @clappr/core

pedrofurtado:
Great idea! @kslimani 👍

pedrofurtado:
@leandromoreira We can help creating PRs 👍 I think the best approach, at least for a while, is to catch possible errors in requestfullscreen

joaopaulovieira:
Some points:

* As @kslimani said, the solution in this PR does not seem to be the best to correct the behavior described in the issue to which it is linked.

* @pedrofurtado can you check if this problem currently occurs in both Clappr Player and Clappr-Core in isolation?

* The @kslimani suggestion of handle the `catch` seems like the best approach to the reported issue, but I think that we could instead of just calling the `requestFullscreen`, return the Promise so can handle catch. What do you think?

Like:

```javascript
export const Fullscreen = {
...
    requestFullscreen: function(el) {
      if (el.requestFullscreen)
        return el.requestFullscreen()
      else if (el.webkitRequestFullscreen)
        return el.webkitRequestFullscreen()
    ...
}
```

Then:

```javascript
Fullscreen.requestFullscreen(element).catch(error => /* Handle the error */)
```

kslimani:
@joaopaulovieira : returning a Promise seem a good approach. Just keep in mind that some browsers may not return promise. _(like old iOS phone or tablet)_, so ensure it always return a Promise, otherwise the "then" case will attempt to call catch() on null.

pedrofurtado:
Any updates on it? 🤓

joaopaulovieira:
Hi! I opened a [PR on @clappr/core](https://github.com/clappr/clappr-core/pull/38) which I think solves the issue #1860. Feel free to discuss the solution on the PR (:

joaopaulovieira:
Hello! A fix for this behavior is already available in the latest version of Clappr. Closing this PR for this reason and if this problem appears again, new PRs/Issues can be opened (:

PR #2014, opened 2021-05-24
docs(readme): Add Q&A answer for auto play

leaofelipe:
## Summary 

Due to a list of problems related to the autoplay feature, we're adding this new section to the docs.

## Changes

- **readme.md**: Adding new section about autoplay


kslimani:
@leaofelipe i think autoplay feature could be a new chapter in readme table of content, instead of Q&A sub-section.
_(titled, for example , "About autoplay", maybe after "Supported Format" ?)._

I would not use the terms "not working", because it is now an expected behaviour in web browsers.
_(or maybe change it by "not allowed" ?)_

leaofelipe:
It makes a lot of sense @kslimani, I'll work on this 👍

leandromoreira (approved):

leticiafernandes (approved):

PR #2054, opened 2022-02-17
Replace deprecated String.prototype.substr()

CommanderRoot:
[String.prototype.substr()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr) is deprecated so we replace it with [slice()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/slice) which works similarily but isn't deprecated.

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -99,7 +99,7 @@ player.attachTo(playerElement);
     window.onload = function () {
 
       if (window.location.hash) {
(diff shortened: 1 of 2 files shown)


CommanderRoot:
The PR is ready for review

pedrochamberlain (approved):

leaofelipe (approved):

PR #2118, opened 2023-09-02
Create Monorepo Structure

felipecesr:
- Create a packages folder for workspaces
- Move Clappr to the packages folder
- Add Lerna and Yarn Workspaces to the project


Drevel:
pkg:maven/org.apache.commons/commons-parent@61

package.json
```diff
@@ -1,64 +1,15 @@
...
-    "lint:fix": "npm run lint -- --fix",
-    "release": "MINIMIZE=true rollup --config",
-    "start": "DEV=true rollup --config --watch",
-    "test": "echo 'No tests configured. This project is only for bundling the player components.'",
-    "watch": "rollup --config --watch"
-  },
-  "files": [
-    "/dist",
-    "/src"
+  "name": "clappr-project",
```
arielfurtado, line 2: Here isn't good consider maintain the name clappr?

arielfurtado (changes requested):

jhonatangcavalcanti:
What do you think about keeping the `clappr` name on `packages/clappr` instead of `packages/player`?
Why the `ignore-engines` flag was removed?
What about setting up the scripts like `start` on the root `package.json`?

felipecesr:
@jhonatangcavalcanti why do we need to ignore engines?

jhonatangcavalcanti:
We agree to remove the `ignore-engines` since it is not needed.
We've also documented how to use the `start` script on the mono repo structure.

package.json
jhonatangcavalcanti: We've agreed to keep the `clappr-project` name for the monorepo

PR #2127, opened 2023-10-06
Fix Simple Typo 

VrajPatelGitHub:
There is a small typo in README.md.

Should read behaviour precisely rather than behavior.


ogabrielp:
"Behavior" is the standard American English spelling, so no typos there.

PR #2202, opened 2024-10-03
Improvement in Test Unit Coverage

lucasferreiralimax:
Below are the before and after coverage results

#### 1. HTML Audio Component
**Before Coverage**  
Before HTML Audio Coverage

**After Coverage**  
After HTML Audio Coverage

---

#### 2. HTML Image Component
**Before Coverage**  
Before HTML Image Coverage

**After Coverage**  
After HTML Image Coverage

---

#### 3. Player Component
**Before Coverage**  
Before Player Component Coverage

**After Coverage**  
After Player Component Coverage

---

#### 4. Overall Coverage for All Files
**Before Coverage**  
Before Overall Coverage

**After Coverage**  
After Overall Coverage

---

### DIFF
--- a/packages/clappr-core/src/components/player/player.test.js
+++ b/packages/clappr-core/src/components/player/player.test.js
@@ -65,6 +65,293 @@ describe('Player', function() {
     })
   })
 
+  describe('attachTo', () => {
+    test('attaches the player to a given element', () => {
+      const player = new Player({ source: 'http://video.mp4' })
+      const element = document.createElement('div')
+      player.attachTo(element)
+      expect(player.options.parentElement).toBe(element)
+    })
+  })
+
+  describe('resize', () => {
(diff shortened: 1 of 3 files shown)


lucasferreiralimax:
@jhonatangcavalcanti se puder dar uma olhada e verificar se está tudo certo com a cobertura dos testes, seria ótimo!

leaofelipe:
Hi, @lucasferreiralimax!

Since Player is Clappr's exposure on the page, I think it would be good to adjust the tests to validate the actual feature rather than just the internal methods.

Most of the current tests are only checking if internal methods are being called (e.g., container) instead of validating the complete process. For example, when testing the Player.play() method, instead of verifying if container.play was called, we should check if the video is actually playing at the end of the process. You could observe the DOM to confirm this, among other possible solutions.

PR #2204, opened 2024-10-05
fix: public page header

vander-furtuna:
removed clappr text in header and added png logo, added semantic nav tag for navigation and used flexbox in header to arrange elements on the index from the public route


leaofelipe:
Hi, @vander-furtuna!
At the moment, we don’t have any plans to update our local test page to include your suggestions. Therefore, I'll close your request.

But, anyway, thank you very much for your contribution ❤️

PR #2237, opened 2025-04-29
Adjust core resize event

gugaio:
Adjust the core resize event to use option.width when el.clientWidth is not available. 

The Clappr core resize event was generating events with an ‘undefined’ size on nodejs enviroment. The issue occurs because two methods rely on el.clientWidth, but el is an empty object on node. I made an adjustment to use option.width when el.clientWidth is null, which ensures no impact on web or HTML-based TVs. It’s worth noting that option.width is already used in this event, so this change makes the behavior more consistent.

### DIFF
--- a/packages/clappr-core/package.json
+++ b/packages/clappr-core/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@clappr/core",
-  "version": "0.11.5",
+  "version": "0.11.4",
   "description": "Core components of the extensible media player for the web",
   "main": "./dist/clappr-core.js",
   "module": "./dist/clappr-core.esm.js",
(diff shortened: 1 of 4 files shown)


packages/clappr-core/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@clappr/core",
-  "version": "0.11.4",
+  "version": "0.11.5",
```
pedrochamberlain, line 3: I think it'd be best if we bumped `clappr/core`'s version after merging this MR.

packages/clappr-core/src/components/core/core.js
```diff
@@ -254,7 +256,9 @@ export default class Core extends UIObject {
     const orientation = (window.innerWidth > window.innerHeight) ? 'landscape' : 'portrait'
     if (this._screenOrientation === orientation) return
     this._screenOrientation = orientation
-    this.triggerResize({ width: this.el.clientWidth, height: this.el.clientHeight })
+    const width = this.el.clientWidth || this.options.width;
+    const height = this.el.clientHeight || this.options.height;
```
pedrochamberlain, line 260: Could you remove the semicolons? I believe they're causing a lint error.

pedrochamberlain (changes requested):

pedrochamberlain (approved):

pedrochamberlain:
Fixed on another merge request: https://github.com/clappr/clappr/pull/2239

